PR #53, opened 2026-03-25
feat: add optional unauthenticated metrics listener support

todpunk:
## Summary
Adds backend.metricsListener.enabled/port values (default: disabled, port 9091) to expose a dedicated unauthenticated Prometheus scrape endpoint via METRICS_PORT env var on the backend.

When enabled:
- A named 'metrics' containerPort and Service port are added
- METRICS_PORT env var is injected into the backend container
- The ServiceMonitor scrapes the 'metrics' port at /metrics instead of the authenticated /api/v1/admin/metrics on the http port

When disabled (default), behavior is unchanged.

Breaking change: When metricsListener is disabled, the ServiceMonitor now scrapes /api/v1/admin/metrics on the HTTP port instead of /metrics, which is the correct endpoint for the authenticated API and may require updating any existing Prometheus alerting rules that reference the old path, though the old path would not have got data at before this PR
  
Resolution for https://github.com/artifact-keeper/artifact-keeper/issues/570

Pair PR to https://github.com/artifact-keeper/artifact-keeper/pull/571

## Test Checklist
- [X] Helm template renders without errors

## Infrastructure
- [X] Helm: `helm template` renders correctly


todpunk:
@brandonrc I don't think this repo has as much setup for review requests or CI, so in case you rely on notifications, tagging you directly on this companion PR to the other one we merged.

brandonrc (commented):
Two things to flag:

**1. Default `metricsListener.port` collides with `grpcPort`**

Both default to `9090`. If someone enables `metricsListener.enabled: true` without changing the port, the container will try to bind two listeners to the same port. Consider defaulting `metricsListener.port` to `9091` so it works out of the box.

**2. ServiceMonitor path change for existing users**

The existing ServiceMonitor scrapes `port: http` at `path: /metrics`. This PR changes the fallback (when `metricsListener` is disabled) to `path: /api/v1/admin/metrics`. That's probably the correct path, but it's a behavior change for anyone already using the ServiceMonitor. Worth calling out in the PR description as a fix so it doesn't surprise anyone on upgrade.

charts/artifact-keeper/values.yaml
```diff
@@ -90,6 +90,16 @@ backend:
...
+  # -- Unauthenticated Prometheus metrics listener.
+  # When enabled, the backend starts a second TCP listener on `metricsPort`
+  # serving only `GET /metrics` with no authentication. Intended for
+  # Prometheus scrapers that cannot present credentials. Disabled by default.
+  #
+  # NOTE: if gRPC is also enabled, ensure `metricsPort` differs from
+  # `service.grpcPort` (default 9090).
+  metricsListener:
+    enabled: false
+    port: 9090
```
brandonrc: This port conficts with the grpcPort

todpunk: moved to 9091

todpunk:
That should resolve both feedback items.

todpunk:
@brandonrc is this merge-able? Do we have a release cadence on this repo? Docs are more sparse in this one, obviously, so I have no expectations, but no knowledge either.

brandonrc:
> @brandonrc is this merge-able? Do we have a release cadence on this repo? Docs are more sparse in this one, obviously, so I have no expectations, but no knowledge either.

Totally sorry about the delay, I was not showing IaC love.

PR #55, opened 2026-03-27
fix(security): remove hardcoded credentials, harden Helm defaults

brandonrc:
## Summary
- Remove all hardcoded credential defaults from values.yaml (ADMIN_PASSWORD, jwtSecret, s3 keys, postgres password, meilisearch masterKey, dependencyTrack adminPassword)
- Add `validateSecrets` helper that fails `helm install` if jwtSecret, postgres password, or meilisearch masterKey are empty (when their services are enabled and externalSecrets is not)
- Meilisearch master key now sourced via secretKeyRef instead of plaintext env
- Remove plaintext secrets from mesh-test ArgoCD ApplicationSet parameters (dead generator params cleaned up)
- Enable NetworkPolicy and TLS by default

**Breaking change**: Deployments using default values without `--set secrets.jwtSecret=...` (and other required credentials) will now fail at install time with a clear error. This is intentional: the previous behavior silently used well-known credentials.

## Test Checklist
- [ ] `helm template` with externalSecrets.enabled=true renders correctly
- [ ] `helm template` without jwtSecret fails with clear error
- [ ] `helm template` without postgres.auth.password fails with clear error
- [ ] `helm template` without meilisearch.masterKey fails with clear error
- [ ] Meilisearch pod starts and connects using secretKeyRef

## Infrastructure
- [x] Helm: `helm template` renders correctly (with required secrets set)
- [ ] Helm: Manually verified on staging cluster (if applicable)
- [x] ArgoCD: Application manifests are valid


JounQin:
Hi, @brandonrc

It seems this change makes the [Quick Start](https://github.com/artifact-keeper/artifact-keeper-iac#quick-start) invalid now, the related document should also be updated?

brandonrc:
> Hi, @brandonrc
> 
> It seems this change makes the [Quick Start](https://github.com/artifact-keeper/artifact-keeper-iac#quick-start) invalid now, the related document should also be updated?

Yes! 

I am spending some time right now workinh on tests and validation. To bring the baseline up for regression testing and reliability.  

Docs are on the list to follow. 

I will make sure to work those into the backlog!

I don't want something quickstart to fail for someone.

PR #78, opened 2026-04-26
feat(chart): make appVersion load-bearing for image tags

brandonrc:
## Summary

Replaces #77, which proposed an alias-tag mechanism inside `chart-releaser` config. Senior review (Backend Architect + DevOps Automator) recommended dropping that approach and instead making `appVersion` itself the load-bearing field that drives the rendered image tag.

This PR does exactly that:

- `templates/{backend,edge,web}-deployment.yaml` render the image as
  `"{{ .Values.x.image.repository }}:{{ .Values.x.image.tag | default .Chart.AppVersion }}"`.
- `values.yaml` sets `backend.image.tag: ""`, `web.image.tag: ""`, and `edge.image.tag: ""` so the default fires for normal chart consumers. `pullPolicy` flipped from `Always` to `IfNotPresent` since the rendered tag is now immutable per chart version.
- `values-staging.yaml` explicitly sets `web.image.tag: dev` to preserve the previous staging behavior (the staging overlay only overrode the backend tag, so before this PR `web` was implicitly inheriting `dev` from the default).
- `RELEASING.md` documents the release flow, the appVersion to image tag contract, and how consumers discover new versions (no automatic alias).

Result: `helm install ak --version 1.2.0` always pulls image tag `1.2.0`. No out-of-band metadata, no alias step. The image-reference verification gate added in #76 catches mismatches because the rendered tag is now driven by `appVersion`.

## Verification

```
$ helm template foo charts/artifact-keeper --set postgres.auth.password=x --set secrets.jwtSecret=y --set edge.enabled=true | grep -E 'image:.*artifact-keeper'
          image: "ghcr.io/artifact-keeper/artifact-keeper-backend:1.1.0"
          image: "ghcr.io/artifact-keeper/artifact-keeper-edge:1.1.0"
          image: "ghcr.io/artifact-keeper/artifact-keeper-web:1.1.0"

$ helm template foo charts/artifact-keeper -f values-staging.yaml ... | grep artifact-keeper
          image: "ghcr.io/artifact-keeper/artifact-keeper-backend:dev"
          image: "ghcr.io/artifact-keeper/artifact-keeper-web:dev"

$ helm template foo charts/artifact-keeper -f values-production.yaml ... | grep artifact-keeper
          image: "ghcr.io/artifact-keeper/artifact-keeper-backend:latest"
          image: "ghcr.io/artifact-keeper/artifact-keeper-web:latest"
```

## Test Checklist
- [x] Helm template renders without errors
[...]

PR #110, opened 2026-05-20
fix(dtrack): source DB password from externalDatabase.existingSecret

### LINKED ISSUE #150: dtrack pod CreateContainerConfigError when postgres.enabled=false + externalDatabase.existingSecret (POSTGRES_PASSWORD key absent)
When `postgres.enabled=false` and `externalDatabase.existingSecret` is set, `templates/secrets.yaml` intentionally skips writing `POSTGRES_PASSWORD` into the chart-managed Secret (the operator owns the credential). But `templates/dtrack-deployment.yaml` always references `<release>-secrets:POSTGRES_PASSWORD`, so the dtrack pod fails to start with `CreateContainerConfigError` because the key it points at does not exist.

dtrack should source `ALPINE_DATABASE_PASSWORD` from the operator-provided `externalDatabase.existingSecret` in that branch (e.g. via a new `externalDatabase.existingPasswordKey` value, default `POSTGRES_PASSWORD`), while leaving the in-cluster-postgres and inline-password paths unchanged.

Reproducer and fix in the linked PR.

dragonpaw:
## Summary

When `postgres.enabled=false` and `externalDatabase.existingSecret` is set, `templates/secrets.yaml` intentionally skips writing `POSTGRES_PASSWORD` into the chart-managed Secret (the operator owns the credential). dtrack, however, always referenced `<release>-secrets:POSTGRES_PASSWORD`, so the pod fails to start with `CreateContainerConfigError` because the key it points at does not exist.

This PR makes dtrack source `ALPINE_DATABASE_PASSWORD` from the operator-provided `externalDatabase.existingSecret` in that branch, using a new `externalDatabase.existingPasswordKey` value (default `"POSTGRES_PASSWORD"`). All other paths (in-cluster postgres, or external DB with inline `password`) keep their previous source, so the change is backwards compatible — `helm template` output is byte-identical for those cases.

Reproducer (before this PR):

```bash
helm template ak charts/artifact-keeper \
  --set postgres.enabled=false \
  --set externalDatabase.existingSecret=mysecret \
  --set externalDatabase.host=h --set externalDatabase.port=5432 --set externalDatabase.username=u \
  --set secrets.jwtSecret=jwt --set secrets.s3AccessKey=ak --set secrets.s3SecretKey=sk \
  --set opensearch.auth.password=op \
  --set dependencyTrack.enabled=true --set dependencyTrack.adminPassword=adminpw
```

The rendered dtrack pod env points at `ak-artifact-keeper-secrets:POSTGRES_PASSWORD`, but the rendered Secret has no such key.

## Test Checklist
- [x] Helm template renders without errors (`helm lint` clean)
- [ ] Terraform validates/plans cleanly  *(N/A — chart-only change)*
- [x] Manually verified on staging cluster (CloudSQL + GKE)
- [x] Rollback strategy documented *(pure additive value; revert is safe)*

## Infrastructure
- [x] Helm: `helm template` renders correctly across four value combinations (defaults; external DB inline password; external DB `existingSecret`; external DB `existingSecret` + custom `existingPasswordKey`)

---
Closes #150


brandonrc:
Thanks for this. Now that several chart PRs (including the v1.2.0 alignment #145 and the load-bearing image-tag change #78) have landed on `main`, this branch has merge conflicts. Could you rebase onto the latest `main`, resolve them, run `cd charts/artifact-keeper && helm-docs` if values changed, and push? CI will re-run after.

(The red SonarCloud check is a known non-blocking fork limitation and can be ignored.)

dragonpaw:
Rebased onto the latest `main` as requested — the conflicts are resolved and `helm-docs` (v1.14.2) was regenerated for the new `externalDatabase.existingPasswordKey` value. Also filed and linked a tracking issue (#150, `Closes #150` footer) so the require-linked-issue check is green. `helm lint`/`helm template` clean. Ready for another look.

brandonrc:
Thanks for this. The bug is real and still on main: with an external database and `externalDatabase.existingSecret`, Dependency-Track reads `POSTGRES_PASSWORD` from a Secret the chart never writes.

As written, though, this switches the password source for **every** external-database install. With External Secrets or `secrets.existingSecret`, the chart's own Secret already carries `POSTGRES_PASSWORD`. The production values would then read from `artifact-keeper-rds`, which only holds `DATABASE_URL`, and break Dependency-Track on upgrade.

Here is what I'll push, keeping your commit: take the password from `externalDatabase.existingSecret` only when the chart does not create the Secret itself, or when a password key for that Secret is set explicitly. I'll add render tests for all three setups, then rebase onto main after #339. Every chart PR has to bump `Chart.yaml` `version` past main (1.9.12 today), so I will set the final version and regenerate the README with helm-docs 1.14.2 when this reaches the front of the merge queue. Your commits and authorship stay.

brandonrc:
Correction to my plan above: I had the condition backwards. The password comes from `externalDatabase.existingSecret` when the chart **does** create its own Secret, because that Secret omits `POSTGRES_PASSWORD` for an external database. It also comes from there whenever `externalDatabase.existingPasswordKey` is set explicitly. With External Secrets or `secrets.existingSecret`, the Secret already carries `POSTGRES_PASSWORD`, so behaviour stays as it is on main. The production values keep reading `artifact-keeper-secrets`.

PR #113, opened 2026-05-20
fix(dtrack): source JDBC host/port from externalDatabase.existingSecret

### LINKED ISSUE #151: dtrack JDBC URL has an empty host when using externalDatabase.existingSecret -> Connection refused
On the canonical external-DB path (`postgres.enabled=false` + `externalDatabase.existingSecret`, with the plain `host`/`port` values omitted), `templates/dtrack-deployment.yaml` builds `ALPINE_DATABASE_URL` from the plain-string `externalDatabase.host`/`externalDatabase.port` values, rendering `jdbc:postgresql://:5432/dependency_track`. JDBC interprets the empty host as `localhost` and dtrack fails on startup with `Connection refused`.

dtrack should be able to source the JDBC host/port from the operator-provided `externalDatabase.existingSecret` so the canonical existingSecret path works without re-supplying host/port as plain values. Reproducer and fix in the linked PR.

dragonpaw:
## Summary

When `postgres.enabled=false` and `externalDatabase.existingSecret` is set (the canonical external-DB path), `templates/dtrack-deployment.yaml` builds `ALPINE_DATABASE_URL` from `externalDatabase.host` / `externalDatabase.port` plain-string values. Operators following the `existingSecret` pattern who omit those plain values get an empty-host JDBC URL — `jdbc:postgresql://:5432/dependency_track` — which JDBC interprets as `localhost`, and dtrack then fails on startup with `Connection refused`.

This PR projects host/port out of the operator-provided Secret into two leading env vars (`_DTRACK_DB_HOST`, `_DTRACK_DB_PORT`) and references them from `ALPINE_DATABASE_URL` via Kubernetes `$(VAR)` expansion. Two new values let operators point at custom keys in the Secret:

```
externalDatabase.existingHostKey  (default "POSTGRES_HOST")
externalDatabase.existingPortKey  (default "POSTGRES_PORT")
```

Mirrors the precedent in #110 (`existingPasswordKey`). Strictly additive: when `existingSecret` is unset, the literal host/port substitution is preserved — `helm template` output is byte-identical for the in-cluster-postgres and inline-host/port paths.

A longer-term alternative would be to consume the full `DATABASE_URL` the backend already reads from the Secret (parse `postgresql://...` → `jdbc:postgresql://...` in the template). Happy to do that follow-up if you'd prefer a single source of truth across both deployments.

Reproducer (before this PR):

```bash
helm template ak charts/artifact-keeper \
  --set postgres.enabled=false \
  --set externalDatabase.existingSecret=mysecret \
  --set externalDatabase.username=u \
  --set secrets.jwtSecret=jwt --set secrets.s3AccessKey=ak --set secrets.s3SecretKey=sk \
  --set opensearch.auth.password=op \
  --set dependencyTrack.enabled=true --set dependencyTrack.adminPassword=adminpw
```

Renders `ALPINE_DATABASE_URL: "jdbc:postgresql://:5432/dependency_track"`. After this PR the same command renders the URL with `$(VAR)` expansion against the two new env vars sourced from `mysecret`.

## Test Checklist
- [x] Helm template renders without errors (`helm lint` clean)
- [x] Manually verified renders across four value combinations
- [x] Rollback strategy documented *(pure additive values; revert is safe)*

## Infrastructure
- [x] Helm: `helm template` renders correctly: defaults / inline external DB / `existingSecret` only / `existingSecret` + custom host/port keys
- [x] Helm: `helm-docs` regenerated (README values table updated)

---
Closes #151


brandonrc:
Thanks for this. Now that several chart PRs (including the v1.2.0 alignment #145 and the load-bearing image-tag change #78) have landed on `main`, this branch has merge conflicts. Could you rebase onto the latest `main`, resolve them, run `cd charts/artifact-keeper && helm-docs` if values changed, and push? CI will re-run after.

(The red SonarCloud check is a known non-blocking fork limitation and can be ignored.)

dragonpaw:
Rebased onto the latest `main` as requested — conflicts resolved and `helm-docs` regenerated for the new `externalDatabase.existingHostKey`/`existingPortKey` values. Filed and linked a tracking issue (#151) so require-linked-issue passes. Re-verified the render: with `externalDatabase.existingSecret` set, `ALPINE_DATABASE_URL` now expands `$(_DTRACK_DB_HOST):$(_DTRACK_DB_PORT)` from the Secret; default paths render byte-identical. Ready for another look.

brandonrc:
Thanks for this. The bug is real: `values-production.yaml` leaves `externalDatabase.host` empty, so main renders `jdbc:postgresql://:5432/dependency_track`.

As written, the host and port always come from Secret keys and `externalDatabase.host` is ignored. An install that sets the host in values and keeps only `DATABASE_URL` in its Secret would stop starting.

Here is what I'll push, keeping your commit: use `externalDatabase.host` and `port` when they are set, and fall back to the Secret keys otherwise. I'll add render tests, then rebase after #110. Every chart PR has to bump `Chart.yaml` `version` past main (1.9.12 today), so I will set the final version and regenerate the README with helm-docs 1.14.2 when this reaches the front of the merge queue. Your commits and authorship stay.

PR #115, opened 2026-05-20
fix(scheduling): auto-inject backend↔trivy podAffinity for RWO scan-workspace PVC

dragonpaw:
## Summary

The `<release>-scan-workspace` PVC is provisioned with `accessModes: [ReadWriteOnce]` but is mounted by **two** Deployments: `backend` (RW) and `trivy` (RO). On multi-node clusters the scheduler regularly places the two pods on different nodes, and the second pod hangs in `Pending` forever with `FailedAttachVolume: Multi-Attach error for volume "...-scan-workspace"`. Single-node clusters work by luck.

This PR auto-injects a required `podAffinity` rule on the backend Deployment that co-schedules it with trivy on the same node (`topologyKey: kubernetes.io/hostname`) whenever **all** of these are true:

1. `backend.affinity` and `global.affinity` are at their default empty value
2. `backend.scanWorkspace.enabled` is `true`
3. `trivy.enabled` is `true`

The auto-injected rule uses the existing `artifact-keeper.trivy.selectorLabels` helper (instance-scoped) so it won't match trivy pods from a different release in a shared namespace.

**Escape hatch:** setting `backend.affinity` (or `global.affinity`) to any non-empty value renders the operator's value verbatim and skips the auto-inject. This is a deliberate all-or-nothing handoff — we don't try to merge auto-inject rules into operator-supplied affinity, because doing so silently would surprise operators who think they're fully owning scheduling. Operators who want to combine podAntiAffinity (to spread backend replicas) with the trivy podAffinity must include both in their own `backend.affinity` block; the new comment on `backend.affinity` in `values.yaml` calls this out.

## Behavior change

Default installs that did **not** previously set `backend.affinity` will now get the new auto-injected `podAffinity` block. On single-node clusters this is a no-op (the constraint is already satisfied). On multi-node clusters this formalizes a constraint that was previously implicit — fresh installs no longer rely on the scheduler happening to co-locate backend and trivy.

The prior default behavior was "works by luck for single-node clusters; deadlocks on multi-node when trivy and backend land on different nodes". Formalizing the constraint surfaces the real requirement.

## Test Checklist
- [x] Manually verified on staging cluster (dev GKE deploy uses an explicit `backend.affinity.podAffinity` of the same shape; this PR auto-injects the same block when operator-affinity is empty)
- [x] Rollback strategy documented *(operators can pin `backend.affinity` to any explicit value — e.g. `{nodeAffinity: ...}` — to skip the auto-inject)*

## Infrastructure
- [x] Helm: `helm template` renders correctly — verified across 5 scenarios (defaults; explicit `backend.affinity`; `scanWorkspace.enabled=false`; `trivy.enabled=false`; `global.affinity` set)
- [x] Helm: `helm lint` clean
- [x] Helm: `helm-docs` regenerated
- [ ] ArgoCD: Application manifests are valid *(no Application changes)*


brandonrc:
Thanks for the PR. This branch has merge conflicts with `main` now, so it cannot be merged as-is. Could you rebase onto the latest `main`, resolve the conflicts, and push? CI will re-run afterward.

Heads up: the red **SonarCloud Scan** is a known non-blocking issue on fork PRs (GitHub withholds `SONAR_TOKEN` from forks; the job is `continue-on-error`), so you can ignore that one.

dragonpaw:
Closing this as superseded. The RWO multi-attach contention this podAffinity worked around came from the `scan-workspace` PVC being mounted by *two* Deployments (backend RW + trivy RO). PR #124 (merged) dropped the dead trivy-side `scan-workspace` mount entirely, so backend is now the sole consumer of that PVC and the cross-Deployment multi-attach risk is gone by construction — no co-scheduling constraint needed. Preferring the #124 approach over pinning the two Deployments together, since it also lifts the implicit single-node restriction. Thanks!

PR #116, opened 2026-05-21
fix(chart): raise backend.resources.limits.ephemeral-storage default to 16Gi

### LINKED ISSUE #153: Backend ephemeral-storage limit of 1Gi evicts the pod on >1 GiB streaming uploads
Kubernetes `ephemeral-storage` accounting includes the container writable layer, logs, and all emptyDir volumes. The backend streams artifacts to the `/data/storage` emptyDir as temp files before finalizing, so the pod's writable usage tracks the full in-flight artifact size. With the default `backend.resources.limits.ephemeral-storage: 1Gi`, any artifact larger than 1 GiB trips kubelet eviction:

```
Pod ephemeral local storage usage exceeds the total limit of containers 1Gi
```

The client sees a TLS EOF mid-stream and the backend logs nothing (the pod was killed before the write completed). The default also disagrees with the chart's own `persistence.size: 10Gi` in the same block. The default limit should be raised (and documented to move together with the emptyDir sizeLimit). Reproduction and validation in the linked PR.

dragonpaw:
## Summary

Bumps the backend pod's `ephemeral-storage` limit from `1Gi` to `16Gi` so streaming format handlers (Incus, large OCI layer pushes) don't get kubelet-evicted on >1 GiB uploads.

## The bug

Kubernetes `ephemeral-storage` accounting includes the container writable layer, container logs, **and** all emptyDir volumes. The chart's `/data/storage` mount is an emptyDir with `sizeLimit: 10Gi` (when `persistence.enabled: false`), and any single artifact streamed to disk by the backend goes into that mount as a temp file before being finalized. Streaming-format handlers like the Incus monolithic + chunked upload paths build the artifact entirely on disk via `tokio::fs` calls (they don't route through `StorageBackend::put_streaming`), so the pod's writable usage tracks the full artifact size in flight.

With the previous `ephemeral-storage: 1Gi` limit, any artifact >1 GiB triggers kubelet eviction:

```
Pod ephemeral local storage usage exceeds the total limit of containers 1Gi
```

The client sees a TLS EOF mid-stream; the backend logs nothing because the pod was killed before the write completed. The deployment recreates the pod and the next retry hits the same wall.

The previous default also disagreed with the chart's own `persistence.size: 10Gi` in the same block — anyone enabling persistence and assuming the limits matched would still hit the eviction.

## Reproduction

Stock-chart install with `STORAGE_BACKEND` set to anything (`filesystem`, `gcs`, `s3`). Create an `incus` repo and PUT a >1 GiB file:

```
HTTP TLS EOF mid-stream
$ kubectl describe pod ak-backend-xxx | grep -A2 Reason:
  Reason: Evicted
  Message: Pod ephemeral local storage usage exceeds the total limit of containers 1Gi.
```

Post-fix: HTTP 201, artifact lands in storage backend, no eviction.

## Validation

Self-hosted dev-shared-gke deploy. 3.4 GiB `.tar.zst` Incus upload:
- Pre-fix: kubelet eviction within ~30s, TLS EOF on client.
- Post-fix: HTTP 201 in 205s @ 17.8 MiB/s, artifact persisted, `kubectl describe pod` shows no eviction events.

## Related

- `artifact-keeper/artifact-keeper#1296` — accept `.tar.zst` Incus filename (also needed to upload our Incus images, but separate fix; this chart change is independently useful for OCI / any large-artifact format).
- `artifact-keeper/artifact-keeper#1297` — incus handler uses a writable staging dir on GCS/S3 backends. Same scenario surfaces both bugs and this PR; could be in any order.

## Note for reviewers

Longer-term, the chart could make this configurable per-format (Incus deployments need more headroom than a deploy serving only npm-sized artifacts), but defaulting closer to `persistence.size` is at minimum required for the chart's own opinionated storage layout to work end-to-end.

---
Closes #153


brandonrc:
Thanks for the PR. The one required check failing here is **Verify helm-docs output**: the chart README needs regenerating after the values change. Could you run this from the repo root and commit the result?

```
cd charts/artifact-keeper && helm-docs
```

(helm-docs v1.14.2, matching `.github/workflows/helm-docs.yml`.) That should clear the check.

Heads up: the red **SonarCloud Scan** is a known non-blocking issue on fork PRs. GitHub withholds `SONAR_TOKEN` from forks and the job is already `continue-on-error`, so it is safe to ignore.

dragonpaw:
Regenerated the chart README via `helm-docs` (v1.14.2) as requested — the Verify helm-docs output check is green now. Also rebased onto the latest `main` and filed/linked a tracking issue (#153) so require-linked-issue passes. Ready for another look.

brandonrc:
Thanks for this. Still needed on main. With persistence off (the object-storage or multi-replica layout), scratch volumes that may grow to about 12Gi sit under a 1Gi limit, so the pod is evicted at 1Gi. Raising only the limit is safe in a patch release.

Here is what I'll push, keeping your commit: fix the `values.yaml` comment that still says "(1Gi by default)", and rebase. Fleet-mode presets still have the same low limits; I'll file that as a follow-up. Every chart PR has to bump `Chart.yaml` `version` past main (1.9.12 today), so I will set the final version and regenerate the README with helm-docs 1.14.2 when this reaches the front of the merge queue. Your commits and authorship stay.

PR #121, opened 2026-05-22
feat(chart): add extraManifests and backend.environmentSecrets

### LINKED ISSUE #120: Helm chart: support extraManifests and backend.environmentSecrets
## Motivation

Two related gaps in the chart make it hard to wire the backend to an authenticated OTLP collector without forking:

1. **No way to source env vars from existing Secrets.** `backend.env` is a plain map rendered as `value:` entries, so anything sensitive (e.g. `OTEL_EXPORTER_OTLP_HEADERS` carrying a bearer token) ends up in plain text in values files / Argo manifests.
2. **No extension point for arbitrary Kubernetes objects.** Operators frequently need to ship an `ExternalSecret`, `SealedSecret`, or custom CR alongside the chart release. Today that requires a separate Application/Kustomize layer.

## Proposal

Add two new values:

### `backend.environmentSecrets`
List of env vars sourced from existing Secrets, mapped into the backend container's `env:` via `secretKeyRef`:

```yaml
backend:
  environmentSecrets:
    - name: OTEL_EXPORTER_OTLP_HEADERS
      secretKeyRef:
        name: otel-collector-auth
        key: otlp-headers
```

### `extraManifests`
Top-level list of raw manifests (templated strings) rendered as additional documents:

```yaml
extraManifests:
  - |
    apiVersion: external-secrets.io/v1beta1
    kind: ExternalSecret
    metadata:
      name: otel-collector-auth
    spec:
      target: { name: otel-collector-auth }
      data:
        - secretKey: otlp-headers
          remoteRef: { key: otel/headers }
```

Each entry is passed through `tpl ... $` so it can reference chart helpers (`{{ include "artifact-keeper.fullname" . }}`).

## End-to-end use case

Together these let an operator deliver an authenticated OTLP exporter config in one chart release: `extraManifests` creates the `ExternalSecret` that materialises `otel-collector-auth`, and `backend.environmentSecrets` injects its value as `OTEL_EXPORTER_OTLP_HEADERS` into the backend pod — no secret material in values, no second Argo Application.

Happy to open a PR.

oba11:
Closes #120

## Summary
- Add `backend.environmentSecrets` — list of env vars sourced from existing Secrets via `secretKeyRef`, rendered into the backend container's `env:`.
- Add top-level `extraManifests` — list of raw, templated manifests rendered as additional documents (`tpl . $`), so chart helpers like `{{ include "artifact-keeper.fullname" . }}` work inside entries.
- Motivating use case: deliver `OTEL_EXPORTER_OTLP_HEADERS` securely by shipping an `ExternalSecret` via `extraManifests` and injecting its value via `backend.environmentSecrets` — no plaintext in values, no second Argo Application.

## Test plan
- [x] `helm lint charts/artifact-keeper` passes.
- [x] `helm template` with both new values set renders the expected `secretKeyRef` entry under the backend container's `env:` and emits the extra manifest as its own document.
- [x] Defaults (`environmentSecrets: []`, `extraManifests: []`) produce no diff vs. main when unset.


oba11:
Hey @brandonrc — quick ping when you have a moment 🙏

All required checks are green. The only red is **SonarCloud Scan**, which fails because `SONAR_TOKEN` isn't exposed to fork PRs.

Happy to tweak anything else. Thanks!

brandonrc:
Thanks for the PR. The one required check failing here is **Verify helm-docs output**: the chart README needs regenerating after the values change. Could you run this from the repo root and commit the result?

```
cd charts/artifact-keeper && helm-docs
```

(helm-docs v1.14.2, matching `.github/workflows/helm-docs.yml`.) That should clear the check.

Heads up: the red **SonarCloud Scan** is a known non-blocking issue on fork PRs. GitHub withholds `SONAR_TOKEN` from forks and the job is already `continue-on-error`, so it is safe to ignore.

oba11:
Thanks @brandonrc, I've regenerated the chart README and all good now.

oba11:
@brandonrc just want to check again on the merge-ability of this pull request. Thanks alot for your help

ThaSami:
@brandonrc i believe this one is quiet good to have, as we are shipping AWS gateway objects and ingresses etc..

PR #162, opened 2026-06-11
feat(chart): expose pod/container/initContainer securityContext per component

### LINKED ISSUE #161: Hardcoded securityContext blocks prevent operators from setting site-specific pod security
### Problem Statement

Every workload in charts/artifact-keeper hardcodes its pod- and container-level securityContext directly in the template. This blocks operators from:                         
                                                                                                                                                                                
  - Setting seccompProfile (required for restricted PSA enforcement in some clusters)                                                                                           
  - Adding supplementalGroups for shared-storage scenarios (NFS, CephFS)                                                                                                        
  - Adjusting runAsUser / fsGroup for environments with reserved UID ranges (e.g. OpenShift's per-namespace UID assignment)                                                     
  - Loosening readOnlyRootFilesystem on the few images that need it for site-local debug builds                                                                                 
                                                                                                                                                                                
  Today the only workaround is ytt/kustomize overlays or post-render patches, which defeat the point of using the chart. We need values-driven knobs.  

### Proposed Solution

  - Each component exposes podSecurityContext, containerSecurityContext, and (where applicable) initContainerSecurityContext.                                                   
  - Defaults preserve current behavior — no diff in helm template output for unchanged installs.                                                                                
  - Init containers with special requirements (e.g. opensearch fix-ownership needing CHOWN/FOWNER as root) get their own dedicated block rather than being lumped with the rest

### Alternatives Considered

_No response_

### Component

Helm Chart

### Additional Context

_No response_

### Pre-submission Checklist

Fixed via #162

nicola-preda:
## Summary
Replaces hardcoded securityContext blocks across every chart workload with values-driven configuration. Defaults match what was previously hardcoded, so helm template output is semantically identical for any unchanged install

### What changed                                                                                                                                                                  

Three new blocks per component in values.yaml (where applicable):                                                                                                             
                                                                                                                                                                                
  - <component>.podSecurityContext                                                                                                                                              
  - <component>.containerSecurityContext                                                                                                                                        
  - <component>.initContainerSecurityContext

Components touched: backend, web, edge, postgres, opensearch, trivy, dependencyTrack (+ the dependencyTrack.bootstrap Job).

### Special cases                                                                                                                                                                 
                                                                                                                                                                                
  - opensearch.fixOwnership — the fix-ownership init container needs runAsUser: 0 + CHOWN/FOWNER, which doesn't fit the shared initContainerSecurityContext shape (used by the  
  sysctl-check sidecar). It gets its own opensearch.fixOwnership.securityContext block, plus a new opensearch.fixOwnership.enabled toggle so operators relying on cluster-side  
  fsGroup chown can skip the init container entirely.                                                                                                                           
  - dependencyTrack.bootstrap — the bootstrap Job gets its own bootstrap.podSecurityContext / bootstrap.containerSecurityContext so it can be hardened independently of the main
   dtrack deployment.                                                                                                                                                           
  - postgres.containerSecurityContext — previously had no container-level securityContext at all; default is {} so render output is unchanged, but the knob is now available.
  
 ### Backward compatibility                                                                                                                                                        
                                                                                                                                                                                
  - helm template against a fresh install with no value overrides produces field-for-field identical output (modulo toYaml's alphabetical key ordering).                        
  - Existing overrides that targeted <component>.podSecurityContext.runAsUser etc. continue to work — the structure is unchanged.                                               
  - One subtle change worth flagging: runAsNonRoot: true was hardcoded in several templates and is now expressed only via the values defaults. An operator who supplies a       
  complete replacement podSecurityContext (rather than patching a single field) and forgets runAsNonRoot will lose the constraint. Documented in the values.yaml comment
  
## Test Checklist

## Infrastructure

Closes #161


brandonrc (commented):
Reviewed the diff and verified rendering. This is a clean, backward-compatible refactor. Nice work.

**Verification I ran:** rendered the chart on `main` and this branch with default values and diffed. The only securityContext-related differences are `toYaml`'s alphabetical key reordering and YAML-equivalent list indentation, exactly as described. Every field (`runAsUser`, `fsGroup`, `runAsNonRoot`, `readOnlyRootFilesystem`, `allowPrivilegeEscalation`, dropped caps, the OpenSearch CHOWN/FOWNER add) is preserved field-for-field for an unchanged install.

A few things I specifically checked that hold up:
- OpenSearch renders both Deployment and StatefulSet from one file via the conditional `kind`, with `securityContext` in the shared pod spec, so the StatefulSet (replicaCount >= 2) path is covered too.
- `sysctl-check` is unconditional, so gating `fix-ownership` behind `fixOwnership.enabled` cannot produce an empty `initContainers:` list.
- The `with` guard correctly omits the `securityContext:` key entirely for `postgres.containerSecurityContext: {}` rather than emitting a dangling empty key.

No correctness issues found. One optional, non-blocking follow-up:

- `postgres.containerSecurityContext: {}` leaves postgres as the only workload with no container-level hardening, while every other component drops `ALL` caps and sets `allowPrivilegeEscalation: false`. This matches prior behavior and the `{}` default is clearly a deliberate choice to keep render output identical, so it is correct as-is. Since the pod already runs as non-root (`runAsUser: 999`), the entrypoint's gosu step-down is skipped, so a later hardening of at least `allowPrivilegeEscalation: false` + `capabilities.drop: [ALL]` would likely work and close the one remaining gap. Worth a separate issue rather than blocking this PR.

Minor note: the "operator supplies a complete replacement podSecurityContext and forgets runAsNonRoot" caveat in the description is largely mitigated by Helm's deep-merge of values (a wholesale map replacement is not actually possible via `-f`/`--set` without explicitly nulling keys), so the real-world risk is smaller than the note implies.

brandonrc (approved):

PR #230, opened 2026-07-15
feat(chart): support CloudNativePG dedicated database per service

### LINKED ISSUE #231: Support a dedicated database per service (CloudNativePG)
## Problem

The chart currently offers two database topologies:

- the bundled in-cluster PostgreSQL (`postgres.enabled=true`), and
- a single shared `externalDatabase` used by both the backend and DependencyTrack.

There is no supported way to give **each service its own dedicated database**. This matters for operators running [CloudNativePG](https://cloudnative-pg.io/) (or any managed Postgres), where the natural pattern is one `Cluster` per service, each with its own auto-generated `<cluster>-app` Secret (`uri`, `jdbc-uri`, `host`, `port`, `username`, `password`). Today you cannot point the backend and DependencyTrack at two separate Secrets.

## Proposal

Add a per-service database mode where the backend and DependencyTrack each read their connection details from their own existing Secret:

- `backend.database.existingSecret` and `dependencyTrack.database.existingSecret`
- configurable `secretKeys` that default to the CloudNativePG application-secret key layout and can be remapped (e.g. `fqdn-uri` for cross-namespace clusters)
- activated when `postgres.enabled=false` **and** `externalDatabase.enabled=false`

Requirements:

- **Backward compatible** — the existing shared `postgres` and legacy `externalDatabase` paths must be unchanged when the new mode is not enabled.
- The `wait-for-postgres` init containers must target the correct per-service host.
- The shared Secret must not emit `POSTGRES_PASSWORD`/`DATABASE_URL` when a dedicated database is in use.
- Ship a documented `values-cnpg.yaml` profile plus guardrails (required `existingSecret` references in the new mode).

## Related

Adjacent DependencyTrack external-database pain points: #150, #151, #152.

mehr74uptempo:
Closes #231

## Summary

Add a per-service database mode so the backend and DependencyTrack can each point at their own PostgreSQL instance (e.g. a CloudNativePG `<cluster>-app` Secret) instead of sharing the in-cluster postgres.

The mode activates when `postgres.enabled=false` and `externalDatabase.enabled=false`. Each service reads its connection details from its own existing Secret via configurable key names that default to the CloudNativePG application-secret layout (uri/jdbc-uri/host/port/username/password) and can be remapped (e.g. to `fqdn-uri` for cross-namespace clusters).

Details:
- New `artifact-keeper.dedicatedDatabase` helper plus per-service secret-key lookup helpers in _helpers.tpl.
- backend: DATABASE_URL and the wait-for-postgres init container (PGHOST/PGPORT/PGUSER) sourced from `backend.database.existingSecret`.
- DependencyTrack: ALPINE_DATABASE_URL/USERNAME/PASSWORD and its wait-for-postgres init container sourced from `dependencyTrack.database.existingSecret`.
- secrets.yaml no longer writes the shared POSTGRES_PASSWORD/DATABASE_URL when a dedicated database is in use.
- New `values-cnpg.yaml` profile documenting the two CNPG Cluster resources and the overlay usage.
- DependencyTrack bootstrap Job now uses a configurable image that ships curl+jq (default badouralix/curl-jq) instead of `apk add` at runtime, which failed under the non-root pod security context.
- Added guardrails: minimum 32-char jwtSecret, required existingSecret references in dedicated mode, and a required DependencyTrack adminPassword when the bootstrap Job is enabled.

Fully backward compatible: the default shared `postgres` and the legacy `externalDatabase` paths are unchanged when the new mode is not enabled — the new branches are gated behind `artifact-keeper.dedicatedDatabase`, which is only true when both `postgres.enabled` and `externalDatabase.enabled` are false.

## Test Checklist

## Infrastructure

Rendering verified both ways:

```
# dedicated-DB (CNPG) mode
helm template ak charts/artifact-keeper -f charts/artifact-keeper/values.yaml \
  -f charts/artifact-keeper/values-cnpg.yaml \
  --set secrets.jwtSecret=$(openssl rand -base64 48)

# default shared postgres (unchanged)
helm template ak charts/artifact-keeper --set secrets.jwtSecret=$(openssl rand -base64 48)
```


mehr74uptempo:
Pushed a fix for the failing checks.

## Root cause
The new `validateSecrets` guard fails `helm template` when `dependencyTrack.bootstrap.enabled` is `true` but `adminPassword` is empty (`externalSecrets` off, `dependencyTrack` on). That's the shipped default, so the **default** and **staging** variants stopped rendering — which is why `Template & Validate` and `Verify Image References` failed for exactly those two variants. `production` passes because it sets `externalSecrets.enabled=true` (whole secrets template is skipped) and `smoke`/`mesh-*` pass because they set `dependencyTrack.enabled=false`.

## Fix
Defaulted `dependencyTrack.bootstrap.enabled` to `false` in `values.yaml`. The bootstrap Job can't succeed without an admin password anyway, so operators now opt in by enabling it together with `dependencyTrack.adminPassword`. README regenerated with helm-docs. Verified all variants (default, production, staging, smoke, mesh-main, mesh-peer, and the new cnpg profile) render cleanly with no `adminPassword` supplied.

I also dropped the two earlier `helm-ci.yml` commits — since this PR is from a fork, the checks run the base branch's workflow, so those edits had no effect on CI (and one of them was contributing an `adminPassword=ci-test` string to the GitGuardian findings).

> ⚠️ **Behavior change:** `dependencyTrack.bootstrap` is now off by default. Enable it explicitly alongside `dependencyTrack.adminPassword`.

## Remaining checks (not from this diff)
- **SonarCloud** — infra failure: the SonarSource action can't import its GPG key (`gpg: error running '/usr/bin/dirmngr': probably not installed`) on the `ak-ci-runners` runner. Needs a re-run / runner fix.
- **GitGuardian** — the remaining hits are the repo's existing non-prod CI test values (`ci-nonprod-jwt-…`, meilisearch `masterKey`) already present in `helm-ci.yml` on `main`. A fork PR can't dismiss these from the dashboard.

Could a maintainer re-run SonarCloud and clear the GitGuardian test-secret findings? Thanks!

brandonrc (commented):
Thanks for this, @mehr74uptempo. Giving the backend and DependencyTrack each their own PostgreSQL via a CloudNativePG `<cluster>-app` Secret is a capability we want in the chart, and the design here is thoughtful: the per-service `secretKeys` remap, the `fqdn-uri` note for cross-namespace clusters, and the guardrails (jwtSecret length, required `existingSecret` in dedicated mode, required dtrack `adminPassword` when the bootstrap Job runs) all show real care. I rebased the branch locally onto current `main` and rendered both paths to check it; notes below.

## What I verified

- `helm lint` passes, and `helm template` renders cleanly in both the default shared-postgres path and the CNPG path (`-f values-cnpg.yaml`).
- Dedicated mode wires correctly: backend `DATABASE_URL` resolves to `ak-backend-db-app/uri`, the dtrack DB vars resolve to `ak-dtrack-db-app`, and the chart-managed Secret no longer emits the shared `POSTGRES_PASSWORD`/`DATABASE_URL`.
- Backward compatibility holds: because activation is gated on `hasKey .Values.externalDatabase "enabled"` and `values.yaml` ships no `enabled` key, existing installs are untouched unless they opt in. Nice touch.
- The guardrails fire as intended (missing `backend.database.existingSecret` and a sub-32-char `jwtSecret` both fail the render with clear messages).

## Please rebase onto current main (1.7.1)

The branch is based on chart 1.5.8, and `main` has since moved to 1.7.1. The rebase is mostly clean but there is one real conflict plus a couple of things to reconcile:

1. **`templates/dtrack-deployment.yaml`, `ALPINE_DATABASE_PASSWORD` (around line 145).** This is the one hard conflict. 1.7.1 changed the fallback secret name from a hardcoded `{{ include "artifact-keeper.fullname" . }}-secrets` to `{{ include "artifact-keeper.secretName" . }}` so that `secrets.existingSecret` is honored. Please keep your `dedicatedDatabase` branch, but in the `{{- else }}` branch use `secretName`:
   ```
   {{- if include "artifact-keeper.dedicatedDatabase" . }}
   name: {{ .Values.dependencyTrack.database.existingSecret }}
   key: {{ include "artifact-keeper.dtrack.dbSecretKey" (dict "root" $ "key" "password" "default" "password") }}
   {{- else }}
   name: {{ include "artifact-keeper.secretName" . }}
   key: POSTGRES_PASSWORD
   {{- end }}
   ```
   Keeping the old `-secrets` literal here would quietly regress the `secrets.existingSecret` support that landed in 1.7.1.

2. **`templates/backend-deployment.yaml`, `DATABASE_URL` fallback.** Same idea: the `{{- else }}` branch should read from `{{ include "artifact-keeper.secretName" . }}` rather than a `fullname`-derived literal. It merged that way for me automatically, but please confirm after you rebase.

3. **`templates/_helpers.tpl`, `validateSecrets`.** 1.7.1 rewrote this so the whole block short-circuits when `externalSecrets.enabled` OR `secrets.existingSecret` is set (and `secrets.yaml` only calls the helper when neither is set). Your added `{{- if not .Values.externalSecrets.enabled -}}` wrapper around the postgres/opensearch/adminPassword checks becomes redundant against that gate. It is not harmful, but please fold those checks back into the existing `{{- else -}}` block so the helper stays readable. Your new `dedicatedDatabase` `existingSecret` checks are correctly placed outside it.

## Relationship to `fleet.externalDatabaseBootstrap`

1.7.1 already has a `fleet.externalDatabaseBootstrap` path, and it is worth being explicit about how the two relate since both touch "backend gets its database from an existing Secret." They are orthogonal and do not collide by default: fleet mode never sets `externalDatabase.enabled=false`, so `dedicatedDatabase` stays false and the fleet `DATABASE_URL` path (via `secretName`/`secrets.existingSecret`) is unaffected. They do solve different problems, though. The fleet bootstrap path targets one shared PostgreSQL server with one role and database provisioned per release by a hook Job, backend only. This PR gives a single install a dedicated cluster per service, backend and DependencyTrack, and relies on the CNPG operator to provision. Could you add a sentence to `values-cnpg.yaml` (or the chart README) noting the two are mutually exclusive and that dedicated mode takes precedence if both are somehow set? That will save an operator some confusion.

One related nit: `backend.database.secretKeys.uri` (default `uri`) parallels the pre-existing `externalDatabase.existingSecretKey` (default `DATABASE_URL`). The richer multi-key shape is justified here because the wait-for-postgres init container needs `host`/`port`/`username` too, so I am not asking you to collapse them, just flagging the overlap for the docs.

## Bundled changes I would like to see split or called out

The DependencyTrack bootstrap changes are a real fix, but they are tangential to CNPG and change default behavior, so they deserve their own visibility:

- **`dependencyTrack.bootstrap.enabled` flips `true` to `false`.** For an existing install that relied on the bootstrap Job, this stops it from running on the next upgrade. That is defensible given it now hard-requires `adminPassword`, but please call it out as an explicit upgrade note (and ideally split it into its own PR so it is easy to review and revert independently).
- **Bootstrap image `alpine:3.20` + `apk add curl jq` becomes `badouralix/curl-jq:latest`.** Shipping a `:latest` third-party Docker Hub image as the chart default is the part I would push back on. Please pin a specific tag or digest for the default. The chart supports airgap operators (see the edge-image note), and a floating third-party tag makes pre-pull and supply-chain review harder. Your own values comment already recommends pinning for production, so let's make the default reflect that.

## Minor

- `values-cnpg.yaml` line 46 has trailing whitespace on the blank line after `postgres.enabled: false`.
- In dedicated mode the wait-for-postgres init container still uses `postgres.image` (`postgres:16-alpine`) for `pg_isready`, even though the in-cluster postgres is disabled. That is fine and pragmatic, but worth a one-line note so airgap pre-pull lists keep `postgres.image` included.
- Consider adding `enabled: true` to the `externalDatabase` block in `values.yaml` with a short comment. Today the activation key only appears in `values-cnpg.yaml`, so someone reading `values.yaml` has no signal that `externalDatabase.enabled=false` is the switch. Making it a documented key (defaulting to `true` to preserve current behavior) would also let you drop the `hasKey` guard.
- The CNPG app-secret key default `username` looks right for current CloudNativePG, but please double check it against the operator version you are targeting, since the key set has varied across CNPG releases. The `secretKeys` remap covers operators who differ, which is good.

Net: this is close, and it is a capability we want in the chart. After a rebase onto 1.7.1 (the dtrack `secretName` conflict is the only hard one), a decision on the bundled bootstrap image/default, and the small doc additions, I think this is in good shape to merge. Thanks again for the careful work and the rendering evidence in the description.

nicola-preda:
I suggest running the cloudnative-pg operator externally, disabling the helm built-in postgres and using a `Cluster` CRD resource instead of baking-in the postgres. That's how i'm currently doing stuff, an not only for AK

brandonrc:
Thanks for this. The credential wiring is right: the key names match the CloudNativePG app Secret for both backend and Dependency-Track. The chart creates no CNPG resources, so there's no CRD-version risk. Two things block merging:
1. **The required-Secret checks are skipped when `externalSecrets.enabled` is set,** and the production profile uses it. Rendering that combination produced 10 empty `secretKeyRef name:` fields.
2. **The bundled Dependency-Track bootstrap changes would break default installs.** Main fixed the bootstrap Job in #265, and this branch turns it off by default, so no API key is written. It also switches to a third-party `badouralix/curl-jq` image and adds an `adminPassword` check that doesn't fit main's generated password.

As asked in July: please rebase onto main (the branch is still at chart 1.7.5), drop the Dependency-Track bootstrap changes, and make the Secret checks cover the External Secrets path. This PR also edits the same Dependency-Track database env block as #110 and #113, which land first, so rebasing after them will be easiest.

PR #265, opened 2026-07-29
Make the Dependency-Track bootstrap survive a default install

### LINKED ISSUE #202: DependencyTrack bootstrap is fragile: dtrack-init apk-adds curl at runtime, adminPassword can be empty, backend never retries the key
On a fresh helm install with `dependencyTrack.enabled: true`, DT ends up permanently "unavailable" in the UI. Three compounding chart bugs (found on a live v1.3.0 deploy):

1. **`dtrack-init` installs tooling at runtime** — its command is `apk add --no-cache curl jq && /scripts/init-dtrack.sh`, but the whole script depends on curl+jq. On any cluster without pod egress to the Alpine package CDN (common — image pulls go through a mirror, but arbitrary pod egress is often blocked), `apk add` fails silently (`>/dev/null 2>&1`) → `curl: not found` → the readiness wait loop can never succeed → job times out ("timed out waiting for the condition") → no API key is ever written. **Fix: use an init image with curl+jq baked in (e.g. a pinned curl+jq image), not `apk add` at runtime.**
2. **`dependencyTrack.adminPassword` defaults to empty** (values.yaml) → the init would call `forceChangePassword` with an empty new password, which DT rejects (406). **Fix: make it required, or auto-generate a strong password into the secret when empty.**
3. **The backend reads `DEPENDENCY_TRACK_API_KEY_FILE` only once at startup** (no retry) → even after DT + init eventually succeed, DT stays "disabled" until the backend is restarted (the startup race the #199 review flagged). **Fix: retry/re-read the key on a timer, or gate backend start on init completion.**

Worked around live by: setting a DT admin password, bootstrapping DT manually (force-change → login → create Automation API key), writing the key to /shared/dtrack-api-key, and restarting the backend → DT then reports enabled+healthy. Related to #199 (same dtrack-init area). This is a default-deploy usability bug, not a v1.3.0 regression.

brandonrc:
## Summary

Fixes the three chart-side bugs in #202. On a default install with `dependencyTrack.enabled: true` (the chart default), Dependency-Track ended up permanently "unavailable" in the UI, and the only symptom was a Job that sat for five minutes and timed out with no useful message.

**One correction to the issue's diagnosis.** #202 describes the `apk add` failure as conditional on blocked egress. It is not conditional. The bootstrap pod runs as non-root (`bootstrap.podSecurityContext` sets `runAsNonRoot: true, runAsUser: 1000`) and `apk add` requires root, so it failed on every cluster. The `>/dev/null 2>&1` redirect hid it. That makes this the default self-hosting experience rather than an edge case.

### What changed

**No runtime package installs.** The Job now runs on an image that already carries `curl` and installs nothing. The image is exposed as `dependencyTrack.bootstrap.image` so air-gapped operators can point at their own registry, and the Job now honors `global.imagePullSecrets` like the other workloads so that registry can require auth.

**Admin password is generated when unset.** `adminPassword: ""` meant an empty password was sent to `forceChangePassword`, which Dependency-Track rejects with 406. The chart now generates a 32-character password. Stability across renders comes from two mechanisms, because they cover different deployment paths:

- Plain `helm upgrade`: the helper reads the password back out of the chart's own Secret with `lookup`, so it stays stable instead of rotating and locking the bootstrap out of the instance it just configured.
- ArgoCD: `lookup` returns nothing under `helm template`, which is how ArgoCD renders, so on that path every sync would have minted a fresh password. The ApplicationSet now sets `ignoreDifferences` on `DEPENDENCY_TRACK_ADMIN_PASSWORD` with the `RespectIgnoreDifferences=true` sync option, so the live value is kept after first creation. The caveat for other template-mode consumers (Flux and similar) is documented in values.yaml and the helper.

An explicit `adminPassword` still wins, and is now sent with `--data-urlencode` so passwords containing form metacharacters work.

**The Job is a hook now.** `post-install,post-upgrade` with `before-hook-creation,hook-succeeded` deletion, same pattern as the fleet DB bootstrap Job. A Job's pod template is immutable, so as a plain resource this PR's image and command change would have failed every `helm upgrade` on an existing install with a "field is immutable" error. The hook recreates the Job cleanly each time; the script short-circuits when the API key is already provisioned, so re-runs are cheap.

**Errors surface.** `set -eu`, no output suppression, a named failure for each stage, and the readiness timeout prints what `curl` actually reported.

**The `jq` dependency is gone**, which is what lets the image stay minimal. That means extracting one field by hand, so the parse is deliberately narrow and validated rather than trusted: team objects are split onto separate lines so a match cannot run across a boundary and pick up another team's key, and the extracted value must look like an API key before it is written. A malformed extraction fails the deploy loudly instead of writing a bad key that resurfaces later as confusing 401s. This extraction assumes Dependency-Track < 4.13 (4.13 hashes API keys and stops returning them in plaintext); that landmine is now documented next to the pinned image tag.

**Docs and CI.** The chart README gains a runbook for recovering a password mismatch after uninstall/reinstall, the README is regenerated with helm-docs, the chart version is bumped to 1.8.1, and `ct install` now runs with `--upgrade` so main-to-PR chart upgrades are exercised in CI. Fuller DT-enabled install coverage is tracked in #269.

### Tradeoff worth reviewing

Hand-parsing JSON is worse than `jq` in isolation. It is chosen here because the alternative is either a runtime package install (the bug) or a new third-party image dependency in a chart that self-hosters pull. If you would rather take a pinned `curl`+`jq` image, the script change is easy to revert on its own and `dependencyTrack.bootstrap.image` already makes it a values-level decision.

### Not in this PR

- The third bug in #202: the backend reads `DEPENDENCY_TRACK_API_KEY_FILE` once at startup, so even a successful bootstrap leaves DT disabled until a restart. That is a change in the backend repo and is being handled separately.
- The NetworkPolicy ingress-namespace restriction: an earlier revision of this branch added a `networkPolicy.ingressNamespace` value without the template wiring, which made it inert. It is removed here and tracked properly in #268.

## Test Checklist

Rollback: revert the commits and bump the chart version. No resource is renamed. Note that a plain revert restores the old `secrets.yaml`, which templates `adminPassword` verbatim; for installs that used a generated password that overwrites the Secret with an empty string on the next sync and desyncs it from Dependency-Track. If you roll back, first copy the current `DEPENDENCY_TRACK_ADMIN_PASSWORD` out of the Secret and set it explicitly as `dependencyTrack.adminPassword` (or accept re-pairing by hand per the README runbook).

## Infrastructure
- [x] ArgoCD: Application manifests are valid

### Verification performed

- `helm lint` clean.
- Rendered with DT enabled: Job carries the hook annotations, image resolves, command is `/bin/sh /scripts/init-dtrack.sh` with no `apk`, generated password is 32 characters.
- Rendered with an explicit `adminPassword` containing `@`, `&`, and `=`: passed through unchanged and sent urlencoded.
- Rendered with DT disabled: no dtrack Job or ConfigMap.
- Rendered with `global.imagePullSecrets` set: the Job pod spec picks it up.
- `sh -n` and `dash -n` on the rendered script: syntax clean.
- Extraction logic tested against synthetic responses including a decoy `AutomationBackup` team listed before `Automation` (correct key selected, quote-delimited name match does not false-positive) and an empty `apiKeys` array (fails loudly, writes nothing).
- helm-docs regenerated; README drift check should now pass.

Not verified: a live install against a real Dependency-Track. Worth doing before release given this is the default path.

Closes #202


brandonrc:
Note for whoever merges: this auto-closes #202, which documents three bugs. Only the two chart-side ones are fixed here. The third (backend reads the API key file once at startup) is now tracked at artifact-keeper/artifact-keeper#2975.

Practical effect of merging this alone: DT bootstrap goes from never completing to completing successfully, and the backend may need one restart to pick up the key. That is strictly better than the current state, so it does not need to wait on the backend fix.

PR #289, opened 2026-08-02
feat(chart): add backend.environmentSecrets for secret-sourced env vars

brandonrc:
## Summary

Ports the `environmentSecrets` half of #121 onto current main. The community branch is based on Chart 1.7.5 (May); merging it as-is would regress `Chart.yaml` below the released 1.9.4 and the release workflow would skip publishing. `extraManifests` stays with the original PR for the contributor to rebase. Co-authorship credited.

Each entry maps a container env var to an existing Secret key:

```yaml
backend:
  env:
    ADMIN_PASSWORD: null   # remove the plain-text default
  environmentSecrets:
    - name: ADMIN_PASSWORD
      secretKeyRef:
        name: ak-admin-credentials
        key: ADMIN_PASSWORD
```

Motivating use case: multi-instance deployments can provision the initial `ADMIN_PASSWORD` out-of-band via an operator-managed Secret, so installs never depend on the pod-local `admin.password` file (which is per-pod on emptyDir-style storage and can diverge from the database across replicas and restarts).

## Test plan

- `helm template` with the values above: exactly one `ADMIN_PASSWORD` env entry, secret-sourced ✓
- Default render (no `environmentSecrets`) unchanged ✓
- CI: lint + k3d install
- Chart 1.9.5 auto-releases on merge


brandonrc:
Superseded by a fresh PR (history hygiene).

PR #292, opened 2026-08-03
Honor existingSecret for chart-managed deployments

balllder:
The chart documents existing-secret usage but always renders artifact-keeper-secrets and hard-codes that name in backend references.

This adds secrets.existingSecret, skips Secret rendering when set, and uses the selected Secret for backend refs. It preserves externally managed migration/peer keys during Helm and Argo adoption.


balllder:
Closing as superseded: upstream main already contains the existingSecret support and values-driven pod security context needed for this deployment. We have switched our GitOps Application back to upstream artifact-keeper-iac at a7f8037.

PR #297, opened 2026-08-05
feat(chart): add egress proxy support (global + per-component)

### LINKED ISSUE #296: Add egress-proxy support to the Helm chart (global + per-component)
### Problem Statement

Running artifact-keeper behind a corporate egress proxy means hand-injecting
`HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` into each component, via per-service `env`
or post-render (ytt/kustomize) patches. It's repetitive and easy to get wrong (e.g.
forgetting `NO_PROXY`, so in-cluster calls try to route through the proxy).

### Proposed Solution

Native proxy support in the chart:

- `global.proxy` (`httpProxy` / `httpsProxy` / `noProxy`) applied to every component.
- `<component>.proxy` to override, or `proxy.enabled: false` to opt a component out
  (no egress path, keeps internal-only components off the proxy).
- Auto-build `NO_PROXY` from cluster DNS suffixes + the in-chart Service names, so
  operators only add external hosts. Emit both upper- and lower-case forms.
- Default internal-only components (web, opensearch) to `proxy.enabled: false`.

Inert unless a proxy is configured, so no change for existing installs.

### Alternatives Considered

Keep hand-injecting proxy env per component via post-render (ytt/kustomize) patches,
the current workaround. Rejected as repetitive and error-prone.

### Component

Helm Chart

### Additional Context

Implemented in #297.

### Pre-submission Checklist

nicola-preda:
## Summary

Native egress-proxy support so operators behind a corporate proxy no longer hand-inject
`HTTP(S)_PROXY` / `NO_PROXY` per component via env or ytt post-render patches. One setting
replaces the pile of per-service proxy env + post-render patches. Closes #296.

- `global.proxy` (`httpProxy` / `httpsProxy` / `noProxy`) applies to every component.
- `<component>.proxy` overrides it, or `proxy.enabled: false` opts a component out
  entirely (no egress path).
- `NO_PROXY` is auto-built from cluster DNS suffixes + every enabled in-chart Service
  name; operators only add extra external hosts via `noProxy`. Both upper- and
  lower-case forms emitted.
- `web` and `opensearch` default to `proxy.enabled: false` (internal-only).

Notes:
- Fully inert unless a proxy is configured — existing renders unchanged.
- `ci/test-values.yaml` exercises the `global.proxy` path.
- README regenerated via helm-docs.
- Chart version left untouched (release flow owns it).

## Test Checklist

## Infrastructure


brandonrc:
Both reds are known and neither is yours: SonarCloud fails on fork PRs by design, and Image Tag Drift is red on `main` too because the chart still pins backend 1.7.1 while 1.8.1 is published — and 1.9.0 lands shortly, so that gap widens before it closes. Filed separately; it is not a condition on this PR.

Milestoned v1.2.0: `global.proxy` replacing per-service env and post-render patches is a chart feature, and v1.1.10 is carrying operational fixes.

brandonrc:
Thanks for this. The core works. Nothing changes unless a proxy is set, and `NO_PROXY` covers the in-cluster services the backend calls. Three problems need fixing before merge:
1. **Image builder.** It landed on main after you branched (#341). `NO_PROXY` lacks the `-buildkitd` service, and the buildkitd pod gets no proxy env even though it pulls base images from outside.
2. **Duplicate `HTTP_PROXY`.** Setting `backend.env.HTTP_PROXY` as well as the global proxy renders `HTTP_PROXY` twice.
3. **Credentials in plain text.** A proxy URL with credentials is written into the Deployment spec, and there's no Secret-based option.

Also needed: the bare `backend` alias in `NO_PROXY`, and documented `proxy:` keys per component. The docs should also say that a per-component override replaces the global value field by field, and that Dependency-Track needs its `ALPINE_HTTP_PROXY_*` settings rather than the plain variables.

Could you let us know whether you plan to continue? If there's no reply within a week, a maintainer will push these fixes to your branch, keeping your commits.

Issue #299, opened 2026-08-05
chart: Dependency-Track API key is handed to the backend over a ReadWriteOnce PVC, which blocks any multi-replica backend


ThaSami:
### What

When `dependencyTrack.enabled` is true (the default), the DT bootstrap writes its API key to a shared PVC and the backend reads it from there:

- `templates/dtrack-deployment.yaml:225` — bootstrap writes `API_KEY_FILE="/shared/dtrack-api-key"`
- `templates/backend-deployment.yaml:187` — backend gets `DEPENDENCY_TRACK_API_KEY_FILE=/shared/dtrack-api-key`
- `templates/backend-pvc.yaml:57` — the `shared-config` PVC is `accessModes: [ReadWriteOnce]`

An RWO volume can only be attached to one node, so every backend pod has to be co-located with the Dependency-Track pod. Consequences:

1. **`replicaCount > 1` does not work.** Backend pods scheduled onto any other node sit in `ContainerCreating` ("Volume is already exclusively attached to one node"). The `podAntiAffinity` on `kubernetes.io/hostname` in the shipped overlays actively pushes them apart.
2. **Rolling updates are impossible**, since the surge pod hits the same attach conflict. This is the remaining blocker for the `backend.strategy` work in #298 — freeing the other two RWO volumes (`storage` via object storage, `scan-workspace` via a per-pod emptyDir) is not enough while this one is mounted.

### Affected as shipped

- `values-production.yaml`: `backend.replicaCount: 3`, `autoscaling.minReplicas: 3`, `podDisruptionBudget.minAvailable: 2`, `dependencyTrack` enabled — three replicas cannot come up, and under `Recreate` the PDB does not constrain the upgrade anyway.
- `values-staging.yaml`: `backend.replicaCount: 2`, same combination.

### Suggested direction

Deliver the API key through something every pod can read independently rather than a shared filesystem:

- Have the bootstrap Job write the key into a Kubernetes Secret and inject it as `DEPENDENCY_TRACK_API_KEY` (or point `DEPENDENCY_TRACK_API_KEY_FILE` at a projected Secret volume). This drops the PVC from the backend entirely and works with any replica count.
- Failing that, at minimum make the `shared-config` access mode configurable so a ReadWriteMany class can be used, and document that `replicaCount > 1` requires it.

Worth checking against the DT 4.13+ key-hashing landmine already noted in `values.yaml` — a Secret-based hand-off is likely a prerequisite for that rework too.

[...]

PR #300, opened 2026-08-05
feat(chart): make the backend deployment strategy configurable

### LINKED ISSUE #298: chart: backend strategy is hardcoded to Recreate, so every upgrade takes the registry down
### What

`charts/artifact-keeper/templates/backend-deployment.yaml` hardcodes:

```yaml
spec:
  strategy:
    type: Recreate
```

There is no value to override it, so every backend image bump, chart change, or config-driven restart tears all backend pods down before the replacement starts. On a multi-replica install that is a full registry outage on each upgrade — clients see connection failures for the whole pod-restart window, not a drained handover.

### Why it is hardcoded today

The reason is sound: the backend mounts ReadWriteOnce PVCs (`storage`, `scan-workspace`, and `shared-config` when `dependencyTrack.enabled`). A rolling update against an RWO volume leaves the surge pod in `ContainerCreating` with a Multi-Attach error until the rollout times out. So `Recreate` is the right *default* — it just should not be the only option.

### What we would like

A `backend.strategy` value, defaulting to `type: Recreate` so nothing changes for existing installs, that lets installs which hold no RWO volume — artifacts in object storage (`STORAGE_BACKEND=s3`) with `backend.persistence.enabled: false` and `backend.scanWorkspace.enabled: false`, or PVCs on a ReadWriteMany class — opt into a rolling update:

```yaml
backend:
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 0
      maxSurge: 1
```

Note `rollingUpdate` must not be rendered under `type: Recreate` — the Deployment API rejects that combination.

### Related

`backend.scanWorkspace.enabled: false` is currently not usable alongside `trivy.enabled: true`: `SCAN_WORKSPACE_PATH=/scan-workspace` is still set, but no volume is mounted there and the container root filesystem is read-only, so scan writes hit EROFS. Freeing that RWO volume is a precondition for the rolling-update path, so a fix belongs with this.

The third RWO volume (`shared-config`, the Dependency-Track API-key hand-off) is a deeper design issue and is filed separately.

### Context

We hit this on a production artifact-keeper install where the upgrade window is a hard registry outage for every consumer build.

ThaSami:
Closes #298

## Problem

`backend-deployment.yaml` hardcoded `strategy.type: Recreate`, so every backend image bump, chart change, or config-driven restart took the whole backend down before the replacement pod started. On our install that is a hard registry outage in every upgrade window, and there was no value to override it.

## Change

`backend.strategy`, defaulting to `type: Recreate`. Rendered output for **every** shipped values variant (default, production, staging, smoke, mesh-main, mesh-peer, `ci/test-values`) is byte-identical to `main`, so no existing install changes behavior on upgrade. Installs that hold no ReadWriteOnce volume can now opt into a rolling update:

```yaml
backend:
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 0
      maxSurge: 1
```

The strategy body goes through a shared `artifact-keeper.deploymentStrategy` helper that:

- validates the type — `--set backend.strategy.type=rollout` fails the render with `backend.strategy.type="rollout" is not valid; use "Recreate" or "RollingUpdate"`, rather than producing a Deployment the API server rejects;
- only emits `rollingUpdate` under `type: RollingUpdate`, since the Deployment API rejects that block under `Recreate` (so a leftover `rollingUpdate` in values is dropped rather than breaking the install).

Trivy, DependencyTrack, and single-replica OpenSearch deliberately keep their hardcoded `Recreate`: their caches use a single-writer file lock, so two concurrent pods are unsafe whatever the volume access mode is. Only the backend gets the knob.

### Latent fix pulled in: scan-workspace emptyDir fallback

Freeing the backend's RWO volumes is a precondition for the rolling-update path, and one of them could not be turned off at all. With `backend.scanWorkspace.enabled: false` and `trivy.enabled: true`, `SCAN_WORKSPACE_PATH=/scan-workspace` was still injected but no volume was mounted there — and with `readOnlyRootFilesystem: true` every scan write would have failed on EROFS. The scan workspace now falls back to a per-pod `emptyDir` of `scanWorkspace.size`, mirroring how `storage` already falls back when `persistence.enabled` is false. No shipped values file hits that combination today (every overlay that disables `scanWorkspace` also disables `trivy`), so this is latent-bug repair, not a behavior change.

## Known limitation, documented not fixed

The third RWO volume, `shared-config`, is mounted into the backend whenever `dependencyTrack.enabled` is true, because the DT bootstrap passes its API key to the backend through that PVC. That pins every backend pod to the DT pod's node, so DT-enabled installs cannot roll (and cannot run `replicaCount > 1` at all — which is why the shipped `values-production.yaml` `replicaCount: 3` and `values-staging.yaml` `replicaCount: 2` do not work as advertised today). Filed separately as #299 with a suggested Secret-based hand-off.

`values.yaml` and the chart README both spell out all three volumes and which knob frees each, so nobody sets `RollingUpdate` expecting it to work with DT enabled. I deliberately left the values-production/staging overlays alone in this PR — they need #299 first.

## Verification

- `helm lint` clean.
- `helm template` + `kubeconform 0.6.7` (`-strict`, k8s 1.31.0, same flags as `helm-ci.yml`) pass on all six values variants plus `ci/test-values.yaml` and a `RollingUpdate` override with 3 replicas, no persistence, no scan workspace, DT disabled.
- Rendered-manifest diff against `main` is empty for all seven variants (modulo the `helm.sh/chart` version label and the per-render random DT admin password).
- Render matrix on the new value: unset → `type: Recreate`; `strategy: null` → `type: Recreate`; `Recreate` + stray `rollingUpdate` → block dropped; `RollingUpdate` with and without `rollingUpdate` params → correct; invalid type → render fails with the message above.
- `helm-docs 1.14.2` re-run, `README.md` committed (helm-docs drift gate).
- Chart version bumped 1.9.6 → 1.9.7.


ThaSami:
Heads-up on the one red check: **SonarCloud Scan** fails with

```
ERROR Not authorized or project not found. Please check the 'SONAR_TOKEN' environment variable...
EXECUTION FAILURE
```

That is the fork-PR secret restriction, not anything in this diff — GitHub withholds repository secrets from `pull_request` runs on forks, so `secrets.SONAR_TOKEN` is empty (`ci.yml:104`). It is the same failure mode the workflow already documents and skips for dependabot at `ci.yml:84-88`; external contributions hit it too. Extending that `if:` with `github.event.pull_request.head.repo.fork != true` would silence it if you want fork PRs to go green — happy to open that as a separate PR, or leave it to you.

Every other gate is green, including `Verify helm-docs output` and all six `Template & Validate` variants. Install Test (k3d) is still running as of this comment.

oba11:
@brandonrc any chance you could help with prioritizing the review of this PR?

brandonrc:
@oba11 going to work on this now

brandonrc:
Reviewed, verified, and pushed three fixes to the branch. Merging.

## Verification of the claims in the description

I re-ran all of it rather than taking the description on faith, and it holds up:

- **Rendered output is byte-identical to `main`** for all seven shipped variants (default, production, staging, smoke, mesh-main, mesh-peer, `ci/test-values.yaml`), modulo the `helm.sh/chart` label. Confirmed both before and after my fix commit.
- **Render matrix on the new value** behaves as documented: unset -> `type: Recreate`; `strategy: null` -> `Recreate`; `Recreate` plus a stray `rollingUpdate` -> block correctly dropped; `RollingUpdate` with params -> rendered verbatim; invalid type -> render fails with `backend.strategy.type="rollout" is not valid; use "Recreate" or "RollingUpdate"`.
- **The latent scan-workspace bug is real.** On `main`, with `scanWorkspace.enabled: false` and `trivy.enabled: true`, `SCAN_WORKSPACE_PATH=/scan-workspace` is injected with no volume mounted at that path, and `readOnlyRootFilesystem: true` means every scan write hits EROFS. The `emptyDir` fallback fixes it, and the fix is correctly scoped to that combination. Good catch.
- `helm lint` clean, `Install Test (k3d)` green.

## What I pushed (7c68fc0)

- **Regenerated `README.md`** with helm-docs 1.14.2. The per-key `backend.strategy` row was there, but the rolled-up `backend` object row was missing `"strategy":{"type":"Recreate"}`, so the drift gate failed. Almost certainly a casualty of the Aug 16 merge-from-main landing after helm-docs was last run.
- **Bumped the chart 1.9.7 -> 1.9.8.** The description says 1.9.6 -> 1.9.7, but #302 shipped 1.9.7 to `main` on Aug 13, so after the merge-from-main this branch matched `main` and the chart change had no version bump behind it. `Chart Version Check` only emits a warning for that, which is why it was showing green.
- **Dropped two em-dashes** from `README.md.gotmpl` to match the repo writing style, and **documented the ephemeral-storage interaction**: the fallback sets `sizeLimit: 2Gi` from `scanWorkspace.size`, but the backend container's `resources.limits.ephemeral-storage` is `1Gi` by default, so the pod is evicted at the container limit before the emptyDir's own limit ever applies. Anyone flipping `scanWorkspace.enabled: false` needs to raise that too. (The existing `storage` fallback already has this shape at 10Gi vs 1Gi, so it is a pre-existing pattern rather than something this PR introduced.)

## On the two remaining red checks

Neither is caused by this diff and neither blocks merge (there are no required status checks on `main`).

- **Image Tag Drift Check**: `main` itself fails this right now. Backend `v1.7.4` published Aug 14, one day after #302 repointed to `1.7.1`, so every open PR fails the gate. I reproduced it against a pristine `main` worktree to confirm. Repointing here would have shipped a backend upgrade inside a PR whose entire point is byte-identical output, so I used the documented `image-drift-ack` escape hatch and filed the repoint separately as #308. Note the label did not previously exist despite being referenced at `helm-ci.yml:267`, and because the workflow does not trigger on `labeled`, applying it cannot clear an already-red run. Both wrinkles are recorded on #308.
- **SonarCloud**: `SONAR_TOKEN` is withheld from fork PRs, exactly as @ThaSami described. The job is already `continue-on-error: true`, so it is advisory. Extending the existing dependabot `if:` guard to cover forks would stop the recurring red; worth its own PR.

## Follow-ups not addressed here

The knob validates its type but not its preconditions: setting `RollingUpdate` while `persistence.enabled: true` on an RWO class renders fine and then wedges the rollout on Multi-Attach until it times out, and `RollingUpdate` with `replicaCount: 1` still has a gap. The chart cannot know whether a storage class is RWX, which is presumably why this is documented rather than enforced, but a guard requiring an explicit acknowledgement value would turn a stuck production rollout into a render-time error. Worth considering alongside #299.

Leaving `values-production.yaml` (`replicaCount: 3`) and `values-staging.yaml` (`replicaCount: 2`) alone was the right call given `shared-config` pins backend pods to the DT node. #299 is the real blocker for those overlays.

Thanks @ThaSami, and sorry for the wait @oba11.

Issue #312, opened 2026-08-25
Published Helm index is stale at 1.9.7: the gh-pages deployment for the 1.9.8 release failed


kalavt:
### Description

`helm repo add` succeeds against `https://artifact-keeper.github.io/artifact-keeper-iac/`, but the index it serves is stale: it stops at chart 1.9.7 and does not list 1.9.8, so `helm dependency update` / `helm install --version 1.9.8` cannot resolve the current chart through the documented path.

This is a different failure from #51 (which was a 404 — the repository was not published at all). Here the fetch returns 200 and the content is outdated.

### Evidence

The `pages build and deployment` run for the 1.9.8 push failed:

- run `32055372656`, branch `gh-pages`, created `2026-08-17T18:32:11Z` — six seconds after the `artifact-keeper-1.9.8` release was published at `2026-08-17T18:32:05Z`
- job `deploy`, step 2 `Deploy to GitHub Pages`: `failure`

chart-releaser itself worked — the index was committed to the branch, only the Pages deployment of it failed:

| Source | `entries[artifact-keeper]` | `generated` | 1.9.8 listed |
| --- | --- | --- | --- |
| `gh-pages` branch, `index.yaml` | 41 | `2026-08-17T18:32:07` | yes |
| Served site, `index.yaml` | 40 | `2026-08-13T03:56:10` | no |

Both index files point at the same release asset (`releases/download/artifact-keeper-1.9.8/artifact-keeper-1.9.8.tgz`), so the artifact is fine and only its advertisement is missing.

### Steps to Reproduce

```
helm repo add ak https://artifact-keeper.github.io/artifact-keeper-iac/
helm repo update ak
helm search repo ak/artifact-keeper --versions | head -3   # newest is 1.9.7

curl -s https://raw.githubusercontent.com/artifact-keeper/artifact-keeper-iac/gh-pages/index.yaml \
  | grep -c 'version: 1.9.8'                                # 1 — the branch has it
curl -s https://artifact-keeper.github.io/artifact-keeper-iac/index.yaml \
  | grep -c 'version: 1.9.8'                                # 0 — the site does not
```

### Expected Behavior

The published index lists every released chart version, so the documented `helm repo add` path resolves the newest chart.

### Observed Behavior

The published index is three versions of the release process behind, with no error on the consumer side — `helm search repo` simply reports 1.9.7 as newest, which is indistinguishable from 1.9.8 not having been released.

### Additional Context

[...]

PR #315, opened 2026-08-25
ci(release): publish the chart to ghcr.io as an OCI artifact

### LINKED ISSUE #314: Publish the Helm chart to ghcr.io as an OCI artifact alongside the Pages index
### Description

The chart is published only through the GitHub Pages index. When that deployment fails the chart becomes unreachable through every documented path, even though the release asset exists — #312 is a live instance: the 1.9.8 tgz is attached to its release and committed to `gh-pages`, while the served index still advertises 1.9.7.

Publishing the same chart to an OCI registry in the same job removes the single point of failure. The backend, web and scanner-adapter images already publish to `ghcr.io/artifact-keeper/*`, so the registry, the credentials and the org are already in place; only the chart is absent from it. I checked four plausible locations before concluding that, all 404:

```
oci://ghcr.io/artifact-keeper/charts/artifact-keeper
oci://ghcr.io/artifact-keeper/artifact-keeper-iac/artifact-keeper
oci://ghcr.io/artifact-keeper/artifact-keeper/charts/artifact-keeper
oci://ghcr.io/artifact-keeper/helm/artifact-keeper
```

### Expected Behavior

`helm install ak oci://ghcr.io/artifact-keeper/charts/artifact-keeper --version <x.y.z>` resolves any released chart version, independently of Pages.

### Observed Behavior

`helm pull`/`helm install` against any `oci://ghcr.io/artifact-keeper/...` path fails with `NAME_UNKNOWN`, and the Pages index is the only distribution channel.

### Additional Context

Two properties beyond redundancy, both relevant to consumers who pin the chart as a dependency:

- an OCI reference resolves to a digest, so a chart version can be verified and, with cosign, signed the way the images already are
- `dependencies.repository` accepts `oci://`, so a consumer chart can pin the upstream chart without depending on an index being regenerated at all

Happy to send the workflow change; the release job already computes the version it is releasing, so the addition is a package-and-push step plus `packages: write` on that job.

### Pre-submission Checklist

- [x] I have redacted all sensitive information (passwords, tokens, internal hostnames).
- [x] I am using a supported version of the chart or module.

kalavt:
## Summary

Publishes each released chart version to `ghcr.io` as an OCI artifact, in the same release job, alongside the existing Pages index.

Today the Pages index is the only way to reach a released chart, which makes it a single point of failure — and the failure is silent, because a stale index and a current one both answer 200. That is live on 1.9.8 right now (#312): the tgz is attached to its release and committed to `gh-pages`, while the served index still advertises 1.9.7, so `helm repo add` + `helm dependency update` cannot resolve the current chart at all.

The registry, the credentials and the org are already in use — `artifact-keeper-backend`, `artifact-keeper-web` and `artifact-keeper-scanner-adapter` all publish to `ghcr.io/artifact-keeper/*`. Only the chart is absent.

## What changed

`.github/workflows/helm-release.yml`, one step and one permission on the existing `release` job:

- `permissions: packages: write`
- a `Push chart to ghcr.io as an OCI artifact` step after `chart-releaser`, reusing the version the job already derives from `Chart.yaml`

After this, both of these resolve the same version:

```
helm repo add ak https://artifact-keeper.github.io/artifact-keeper-iac/
helm install ak oci://ghcr.io/artifact-keeper/charts/artifact-keeper --version 1.9.8
```

`dependencies.repository` accepts an `oci://` URL, so a consumer chart can pin the upstream chart without depending on an index being regenerated.

## Notes for review

- **The push is not the assertion.** `helm push` says nothing about whether the version became resolvable, so the step pulls the chart metadata back by version (`helm show chart oci://... --version "$version"`) before reporting success. Without that, a release that failed to publish and one that published look identical in the log.
- **The token is passed as an environment variable**, not interpolated into the script body, so it never becomes part of the shell command line.
- **Idempotency.** The step runs under the job's existing `if: needs.version-check.outputs.changed == 'true'`, so it only fires for a version that has no tag yet — matching `skip_existing: "true"` on chart-releaser.
- `$RUNNER_TEMP` is used for the package output rather than the workspace, so nothing lands in the checkout that a later step could pick up.

## Testing

I cannot run this repository's CI, so what I verified locally against `charts/artifact-keeper` at 1.9.8:

- the version expression yields `1.9.8`
- `helm package charts/artifact-keeper --destination <dir>` produces exactly `artifact-keeper-1.9.8.tgz`, the filename the push step references
- the workflow file still parses as YAML, and the `release` job's steps and permissions are as intended

The `helm registry login` / `helm push` / `helm show chart` sequence is untested by me, since it needs `packages: write` in this org. Worth a maintainer eye on the first run — in particular whether `ak-ci-runners` reaches `ghcr.io` outbound.

Closes #314

### DIFF
--- a/.github/workflows/helm-release.yml
+++ b/.github/workflows/helm-release.yml
@@ -79,6 +79,8 @@ jobs:
     runs-on: ak-ci-runners
     permissions:
       contents: write
+      # Pushing the chart to ghcr.io as an OCI artifact.
+      packages: write
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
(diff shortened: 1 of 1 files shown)


brandonrc:
Thanks for this. Checked: `packages: write` is scoped to the release job, the token goes through `--password-stdin`, and the version comes from `Chart.yaml` like the existing check. It only runs when a new chart version is released. I'll squash-merge it. After the first OCI publish, I'll make the ghcr package public.

Issue #313, opened 2026-08-25
shared-config PVC is RWO and unconditional, so dependencyTrack.enabled pins every backend replica to one node


kalavt:
### Description

With `dependencyTrack.enabled: true` the chart mounts `artifact-keeper-shared-config` into the backend, and that PVC is `ReadWriteOnce` with no values knob. On a block-storage class (EBS, GCE PD, Azure Disk) a RWO volume attaches to one node, so every backend replica is pinned to that node and `backend.replicaCount > 1` cannot spread. Enabling the SBOM component therefore costs node-level HA for the whole backend.

The backend does not require the volume. `dependency_track_service.rs` documents that the key may come from the environment and that the direct value wins:

> The API key may instead be supplied via `DEPENDENCY_TRACK_API_KEY_FILE` … the `DEPENDENCY_TRACK_API_KEY` value takes precedence when both are set.

But `templates/backend-deployment.yaml` hardcodes `DEPENDENCY_TRACK_API_KEY_FILE=/shared/dtrack-api-key` and gates the volume only on `dependencyTrack.enabled`, so an operator who supplies `DEPENDENCY_TRACK_API_KEY` through `backend.environmentSecrets` still gets the RWO PVC mounted and still loses multi-node scheduling.

Adjacent to point 3 of #202 (the backend reads the key file once at startup): a Secret-sourced env value removes the file, the init job's write, and that startup race together.

### Chart/Module Version

1.9.8

### Expected Behavior

`dependencyTrack.enabled: true` and a backend spread across nodes can coexist.

### Observed Behavior

The RWO `shared-config` PVC is rendered and mounted unconditionally, and neither its `accessModes` nor its storage class is configurable independently — `templates/backend-pvc.yaml` reads `backend.persistence.storageClass` for it, which is the storage PVC's key and is also consulted when `backend.persistence.enabled` is false.

### Suggested Fix

Any one of these is sufficient; the first is the smallest:

1. Skip the `/shared` volume, its mount and `DEPENDENCY_TRACK_API_KEY_FILE` when the operator supplies the key directly (a `dependencyTrack.existingApiKeySecret` value, or a detected `DEPENDENCY_TRACK_API_KEY` in `backend.env` / `backend.environmentSecrets`).
2. Expose `backend.sharedConfig.{accessModes,storageClass,size}` so the volume can be RWX on a shared filesystem.
[...]

PR #328, opened 2026-09-11
allow separate storage class for shared storage

lboclboc:
Change-Id: Iaae01deba481ed3fabd2fe632ad517d246c8c5d5
Refs #313

## Summary
For the shared storage it might be necessary to use another storage class. In our case we use ceph block storage for most things but for this shared storage that does not work if the dtrack-init pod is started on some other node. Then we need to use ceph-filesystem storage. Also the sharing mode needs to be different.


kalavt:
Flagging one thing before this merges, since the PR body says `Closes #313` and I opened that issue: this implements the second of the three fixes it lists, and the first is the one that removes the constraint rather than relocating it.

**What #313 is about.** With `dependencyTrack.enabled: true`, `templates/backend-deployment.yaml` sets `DEPENDENCY_TRACK_API_KEY_FILE=/shared/dtrack-api-key`, mounts `<release>-shared-config` at `/shared`, and `templates/backend-pvc.yaml` renders that claim — all three gated on `dependencyTrack.enabled` alone. The backend's own key resolution (`resolve_api_key` in `dependency_track_service.rs`, which the chart comment above that env var points at) prefers `DEPENDENCY_TRACK_API_KEY` when it is set. So an operator who supplies the key through `backend.environmentSecrets` still gets the RWO claim attached, and on a block-storage class still gets every replica pinned to one node.

**What this PR changes.** The claim's storage class and access mode become configurable. That is a real improvement, and it is the right answer for an operator who has a shared filesystem to place it on. It does not cover the case in the issue: on a block-storage class no `accessModes`/`storageClassName` pair lets two replicas on two nodes mount the claim, so `backend.replicaCount > 1` stays single-node and the operator has to introduce a shared filesystem they would not otherwise need.

**The half this leaves open** (suggestion 1 in #313): skip the `shared-config` PVC, the `/shared` volume and mount, and `DEPENDENCY_TRACK_API_KEY_FILE` when the key is supplied directly — a `dependencyTrack.existingApiKeySecret` value, or a detected `DEPENDENCY_TRACK_API_KEY` in `backend.env` / `backend.environmentSecrets` — and inject it as a `secretKeyRef` instead. Same shape as `backend.persistence.enabled`, which already does exactly this for `/data/storage`.

Scope note if it is taken up: the `dtrack-init` post-upgrade hook job in `templates/dtrack-deployment.yaml` mounts the same claim and carries no scheduling knob, so the gate has to be applied there too or the upgrade blocker survives the backend-side fix. Measured on 1.9.9 with an external Postgres and a block-storage class: 0 of 4 hook pods landed on the volume's node, and every `helm upgrade` failed with `post-upgrade hooks failed: resource Job/... not ready ... context deadline exceeded`.

**Ask:** this is fine to merge on its own, but please drop `Closes #313` to `Refs #313` so the half that removes the volume stays tracked. Happy to send that PR if it is welcome.

lboclboc:
Does not seem Refs # works, I think am required to use Close #

brandonrc:
Thanks for this. A separate storage class for the shared-config volume is useful. I'll take this over rather than ask for another round:
- **Bug:** `sharedStorageClass` is ignored unless `persistence.storageClass` is also set (`backend-pvc.yaml`).
- **Stale commits:** the branch carries old appVersion and image-tag bumps (1.9.1 / web 1.9.0) that would downgrade main.
- **Linked issue:** there's none, and #313's author asked that it not be closed by this change.

I'll open a follow-up PR that adds the storage class and list-style `accessModes` for shared config, matching #332, credits you as co-author, and links here. I'll close this PR when that one merges.

brandonrc:
Landed via #347, which adds the shared-config storage class (now independent of `persistence.storageClass`) and list-style access modes matching #332. You are credited as co-author. Thanks! Closing this one as superseded.

PR #329, opened 2026-09-11
feat(chart): add backend.podLabels / web.podLabels

### LINKED ISSUE #330: Support backend.podLabels (and web.podLabels) to enable AKS/GKE Workload Identity
### Problem Statement

Both Azure AKS Workload Identity and GKE Workload Identity Federation work the same way: a mutating admission webhook watches for pods carrying a specific label (`azure.workload.identity/use: "true"` on AKS, an equivalent on GKE) whose ServiceAccount carries a matching annotation (`azure.workload.identity/client-id`, `iam.gke.io/gcp-service-account`). When both are present, the webhook injects the federated-token volume, mount, and env vars itself at admission time — the chart doesn't need to render any of that.

`backend.serviceAccount.annotations` already exists and covers the SA side. But there is no way, anywhere in the chart, to add a custom label to the backend (or web) pod template — `template.metadata.labels` in `backend-deployment.yaml`/`web-deployment.yaml` renders only the fixed `*.selectorLabels` helper, with no values-driven passthrough. So the SA annotation is inert: the webhook never fires because the pod never carries the required label.

This is a narrower, separate gap from #305 (`extraVolumes`/`extraVolumeMounts`) — that issue is about mounting arbitrary volumes (e.g. `CUSTOM_CA_CERT_PATH`), which isn't actually required here since the workload-identity webhook injects its own volume independent of the chart. The only missing piece for Workload Identity specifically is the pod label.

### Proposed Solution

Add a generic pass-through, same shape as the existing `tolerations`/`affinity`/`nodeSelector` overrides:

```yaml
backend:
  podLabels: {}
web:
  podLabels: {}
```

rendered into each pod template, e.g. in `backend-deployment.yaml`:

```yaml
  template:
    metadata:
      labels:
        {{- include "artifact-keeper.backend.selectorLabels" . | nindent 8 }}
        {{- with .Values.backend.podLabels }}
        {{- toYaml . | nindent 8 }}
        {{- end }}
```

### Alternatives Considered

Currently the only workaround is a Helm post-renderer or a chart fork — this came up while evaluating Azure Blob Storage as the backend's storage backend on AKS, specifically to authenticate via Workload Identity instead of a shared storage-account key.

### Component

Helm chart (backend + web deployments)

perpective2410:
Both AKS and GKE Workload Identity federation gate their mutating admission webhook on a specific pod label (e.g.
azure.workload.identity/use: "true") in addition to a matching serviceAccount annotation. serviceAccount.annotations already exists on both the backend and web pods, but there was no way to add a custom label to either pod template, so the annotation alone is inert -- the webhook never fires and the projected token volume/mount/env vars it would otherwise inject never show up.

Add a generic podLabels passthrough, in the same style as the existing tolerations/affinity/nodeSelector overrides. Defaults to {}, so behavior is unchanged when unset.

## Test Checklist
- [ ] Helm template renders without errors

## Infrastructure
- [ ] Helm: `helm template` renders correctly

Closes #330


brandonrc:
Thanks for this. The labels land only on the pod template, never on `spec.selector`, so upgrades are safe.

One fix I'll push, keeping your commit: quote the label values in `backend-deployment.yaml` and `web-deployment.yaml`. Your own example, `--set ...azure.workload.identity/use=true`, currently renders a YAML boolean, which the API server rejects. I'll add a render check for it and rebase. Every chart PR has to bump `Chart.yaml` `version` past main (1.9.12 today), so I will set the final version and regenerate the README with helm-docs 1.14.2 when this reaches the front of the merge queue. Your commits and authorship stay.

Issue #305, opened 2026-08-15
Support backend extraVolumes and extraVolumeMounts


imanlotfimahyari:
### Problem Statement

The backend supports configuration that references files on disk, for example
`CUSTOM_CA_CERT_PATH`, but the Helm chart currently has no generic way to add an
additional volume and corresponding volumeMount to the backend container.

This means operators that need to mount an existing ConfigMap, Secret, projected
volume, custom CA bundle, or other configuration file must patch/fork the chart
or use a post-render step.

### Proposed Solution

Add generic pass-through values for additional backend volumes and mounts:

```yaml
backend:
  extraVolumes: []
  extraVolumeMounts: []
```

and render them into the backend Deployment using `toYaml`.

For example:

```yaml
backend:
  extraVolumes:
    - name: custom-ca
      configMap:
        name: internal-ca

  extraVolumeMounts:
    - name: custom-ca
      mountPath: /etc/artifact-keeper/ca.pem
      subPath: ca.pem
      readOnly: true
```

### Component

None

### Additional Context

One concrete use case is `CUSTOM_CA_CERT_PATH`: the backend supports using a
custom CA for outbound HTTPS/OIDC connections, but the chart cannot currently
mount the referenced CA file without an external patch.

Keeping this generic rather than adding a custom-CA-specific volume configuration
would also support projected credentials, additional configuration files, agent
sockets, and similar Kubernetes integrations.

### Pre-submission Checklist

- [x] I have searched existing issues to make sure this is not a duplicate.
- [x] I have reviewed the documentation to confirm this is not already supported.

PR #332, opened 2026-09-14
feat(chart): add backend.persistence.accessModes passthrough

### LINKED ISSUE #331: Support backend.persistence.accessModes to enable ReadWriteMany (Azure Files/NFS) for multi-replica backend
### Problem Statement

`backend.persistence` provisions a PVC for the backend's artifact storage (`templates/backend-pvc.yaml`), but `accessModes` is hardcoded to `ReadWriteOnce`:

```yaml
spec:
  accessModes:
    - ReadWriteOnce
```

On a filesystem-backend deployment, this is the actual artifact store — every replica needs to read/write the same files. A `ReadWriteOnce` PVC (e.g. Azure Disk) zone-pins to whichever node first attaches it, so `backend.replicaCount > 1` just makes every additional replica fight the first for that PV's node affinity (Pending/ContainerCreating), even though `backend.strategy`, `backend.autoscaling`, and `backend.podDisruptionBudget` all already exist and assume real multi-replica is possible.

Cloud environments commonly have a `ReadWriteMany`-capable storage class available (Azure Files, AWS EFS, NFS) specifically for this kind of shared-filesystem use case, but the chart provides no way to select it.

### Proposed Solution

Add a values-driven accessModes list, defaulting to today's behavior:

```yaml
backend:
  persistence:
    accessModes: []  # defaults to ["ReadWriteOnce"] when unset/empty
```

rendered in `backend-pvc.yaml`:

```yaml
spec:
  accessModes:
    {{- range (.Values.backend.persistence.accessModes | default (list "ReadWriteOnce")) }}
    - {{ . }}
    {{- end }}
```

Existing deployments that don't set this see no change (still `ReadWriteOnce`). Setting `accessModes: ["ReadWriteMany"]` with an RWX-capable `storageClass` is then enough to make `replicaCount > 1` + `strategy.type: RollingUpdate` actually schedulable, without moving to an object-storage `STORAGE_BACKEND`.

### Alternatives Considered

`backend.extraVolumes`/`extraVolumeMounts` (#305) would also solve this indirectly (mount an externally-provisioned RWX PVC), but that's a larger, more general passthrough. This is a narrower, one-line fix scoped to the chart's own PVC.

### Component

Helm chart (backend PVC)

perpective2410:
`backend-pvc.yaml` hardcodes `accessModes` to `ReadWriteOnce`, so the backend's artifact-storage PVC always zone-pins to one node -- a second replica just fights the first for that PV's affinity (Pending/ContainerCreating), even though `backend.strategy`, `backend.autoscaling`, and `backend.podDisruptionBudget` all already exist and assume real multi-replica is possible.

Cloud environments commonly have a ReadWriteMany-capable storage class available (Azure Files, AWS EFS, NFS) for exactly this case. This adds a values-driven `accessModes` list, defaulting to today's behavior (`["ReadWriteOnce"]`) so existing installs see no change:

```yaml
backend:
  persistence:
    accessModes: []  # defaults to ["ReadWriteOnce"] when unset/empty
```

rendered as:

```yaml
spec:
  accessModes:
    {{- range (.Values.backend.persistence.accessModes | default (list "ReadWriteOnce")) }}
    - {{ . }}
    {{- end }}
```

Scoped narrowly to the chart's own PVC -- `extraVolumes`/`extraVolumeMounts` (#305) would also solve this indirectly but is a larger, more general passthrough.

## Test Checklist
- [x] Helm template renders without errors (default case: identical output to before, single `- ReadWriteOnce` line)
- [x] Helm template renders correctly with `accessModes: ["ReadWriteMany"]`

Closes #331


brandonrc:
Thanks for this. The change is correct and the default render is unchanged.

Here is what I'll push, keeping your commit:
1. Regenerate the README, which the helm-docs check currently fails on.
2. Add a doc note: this frees only the main storage volume, so running more than one replica also needs `scanWorkspace.enabled=false` and a RWX class for shared config. #328's follow-up will add that.
3. Rebase onto main.

Every chart PR has to bump `Chart.yaml` `version` past main (1.9.12 today), so I will set the final version and regenerate the README with helm-docs 1.14.2 when this reaches the front of the merge queue. Your commits and authorship stay.

PR #333, opened 2026-09-14
dtrack: use same secret for external postgres

### LINKED ISSUE #272: `external-secrets.yaml`: DependencyTrack secret key `DTRACK_ADMIN_PASSWORD` does not match the key consumed by the deployment
# `external-secrets.yaml`: DependencyTrack secret key `DTRACK_ADMIN_PASSWORD` does not match the key consumed by the deployment

## Problem

The ExternalSecret data entry for the DependencyTrack admin password uses:

```yaml
- secretKey: DTRACK_ADMIN_PASSWORD
```

However, `templates/dtrack-deployment.yaml` reads `DEPENDENCY_TRACK_ADMIN_PASSWORD` from the secret,
and `templates/secrets.yaml` (the static secret fallback) also emits it under
`DEPENDENCY_TRACK_ADMIN_PASSWORD`. This mismatch means that when ExternalSecrets is enabled, the
DependencyTrack pod never finds its admin password — it silently reads an empty/missing env var
instead.

## Proposed Change

Align `external-secrets.yaml` with the key name used everywhere else:

```yaml
- secretKey: DTRACK_ADMIN_PASSWORD           # wrong — nothing reads this key
+ secretKey: DEPENDENCY_TRACK_ADMIN_PASSWORD  # matches dtrack-deployment.yaml and secrets.yaml
```

## Why This Matters

Users enabling `externalSecrets` for DependencyTrack will get a silently broken deployment —
DependencyTrack starts but cannot authenticate because the env var it expects is never populated.
The bug only surfaces at runtime, not at render/deploy time.

Happy to provide a PR for this if it would help move it forward.

lboclboc:
Change-Id: I551375671d4feacd236d1abe31c8d24d294d7fcd

## Summary
Adds existingJDBSSecret key to be used by dtrack postgres JDBC connection.
closes #272

## Test Checklist
- [ ] Helm template renders without errors

## Infrastructure
- [ ] Helm: `helm template` renders correctly


brandonrc:
Thanks for this. Closing it because the new condition drops more than intended:
- It also removes `ALPINE_DATA_DIRECTORY` and the database username/password.
- With the in-cluster Postgres plus `externalDatabase.existingSecret`, Dependency-Track would start with no credentials at all.
- The branch bumps unrelated image pins and appVersion without a chart version bump, and links #272, which is closed.

The goal, one Secret for the external database, is being handled in #110 and #113. Your idea of keeping a JDBC URL in that Secret is a good follow-up once those land.

Issue #306, opened 2026-08-15
NetworkPolicy hardcodes ingress-nginx pod labels and blocks other ingress controllers


imanlotfimahyari:
### Description

The Helm chart allows the ingress controller to be selected through `ingress.className`, but the generated NetworkPolicies for the backend, web, and edge components still hardcode the ingress controller pod selector as:

```yaml
podSelector:
  matchLabels:
    app.kubernetes.io/name: ingress-nginx
```

### Component

Helm Chart

### Environment Information

- Kubernetes Version: 1.35
- Helm Version: -
- Terraform Version: -
- Cloud Provider (AWS, GCP, Azure, self-hosted, etc.): EKS
- ArgoCD Version (if applicable): 10.0.1

### Chart/Module Version

1.9.7

### Backend Version

1.7.1

### Expected Behavior

When the chart supports a configurable ingress class, enabling the chart NetworkPolicy should not require the ingress controller to be ingress-nginx.

The ingress-controller pod selector should be configurable so that operators using Traefik or another ingress controller can permit traffic from the actual controller pods without patching the generated NetworkPolicy.

### Observed Behavior

The generated NetworkPolicy uses the configured
`networkPolicy.ingressNamespace`, but still requires the ingress controller pod to have:

```yaml
app.kubernetes.io/name: ingress-nginx
```

### Steps to Reproduce

1. Render or deploy chart version **1.9.7** with **NetworkPolicy** enabled and a non-nginx ingress controller, for example:

   ```yaml
   ingress:
     enabled: true
     className: traefik

   networkPolicy:
     enabled: true
     ingressNamespace: traefik
   ```

2. Inspect the generated backend `NetworkPolicy`.
3. Observe that the ingress rule still contains:

   ```yaml
   podSelector:
     matchLabels:
       app.kubernetes.io/name: ingress-nginx
   ```

4. The same hardcoded selector is also used by the **web** and **edge** `NetworkPolicy` resources.
5. The generated `NetworkPolicy` resources always contain the following selector:

```yaml
podSelector:
  matchLabels:
    app.kubernetes.io/name: ingress-nginx
```
Because Traefik controller pods do not match this selector, traffic originating from Traefik is not allowed by these policies. As a result, ingress traffic from Traefik (or any other non-nginx ingress controller) is blocked.

### Relevant Configuration

```yaml
ingress:
  enabled: true
  className: traefik

networkPolicy:
  enabled: true
  ingressNamespace: traefik
[...]

PR #337, opened 2026-09-16
feat(chart): add optional Gateway API HTTPRoute support

### LINKED ISSUE #322: Chart cannot expose itself via Gateway API — add HTTPRoute support alongside Ingress
### Problem Statement

The chart can only expose itself through a `networking.k8s.io/v1` Ingress. `templates/ingress.yaml` is the only routing template in the chart, and it is hardcoded to `kind: Ingress`.

On clusters that have moved to the Gateway API (Envoy Gateway, Istio, Cilium, Traefik v3, NGINX Gateway Fabric) there is no ingress controller to satisfy that object, so `ingress.enabled: true` renders an Ingress that is simply never reconciled. The install "succeeds" and the workloads run, but nothing is reachable from outside the cluster and there is no error anywhere to explain why.

The workaround is to set `ingress.enabled: false` and hand-write an HTTPRoute. That is more painful than it sounds, because the routing split between the backend and web services is non-trivial and lives entirely inside `ingress.yaml`: **35 path matches** must be reproduced by hand and kept in sync with every chart upgrade.

From `templates/ingress.yaml`:

- `/api` (Prefix), `/v2` (Prefix) → backend
- `/health`, `/ready` (Exact) → backend
- 31 native package-format prefixes → backend:
  `/maven /npm /pypi /nuget /cargo /gems /go /helm /debian /rpm /alpine /composer /conan /conda /swift /terraform /cocoapods /hex /pub /lfs /ivy /chef /puppet /ansible /cran /huggingface /jetbrains /vscode /proto /incus /ext`
- `/dtrack` (Prefix, optional) → dtrack
- `/` (Prefix, catch-all) → web

Getting one of those wrong silently routes a package format to the Next.js frontend instead of the backend, which shows up much later as a confusing 404 from a package client rather than an obvious deployment failure. Any format added upstream in a future chart release also has to be manually mirrored into the hand-written route.

There is prior art in the chart for Gateway-API-adjacent objects — `templates/healthcheck-policies.yaml` renders `networking.gke.io/v1` `HealthCheckPolicy` for GKE Gateway (#125), and #160 added a GKE-Gateway timeout knob. So GKE Gateway users are partially supported today, but only in the sense that the chart emits policies *for* a Gateway someone else created; the chart still cannot emit the route itself.

### Proposed Solution

Add an opt-in `httpRoute` block that renders a `gateway.networking.k8s.io/v1` HTTPRoute with the same path split the Ingress uses, so the two stay in sync from one source of truth.

```yaml
httpRoute:
  enabled: false
  parentRefs:
    - name: envoy
      namespace: classic
      sectionName: artifacts-https   # optional, pin to one listener
  hostnames:
    - artifacts.example.com
  annotations: {}
```

Implementation notes / suggestions:

- Factor the backend path list into a named template (e.g. `artifact-keeper.backendPathPrefixes`) and have both `ingress.yaml` and the new `httproute.yaml` range over it. That is what keeps a future format addition from silently regressing one of the two.
- Fail the render if both `ingress.enabled` and `httpRoute.enabled` are true, or at minimum document that they are mutually exclusive — two objects claiming the same hostname on the same listener conflict.
- TLS needs no chart-side handling: Gateway API terminates TLS at the Gateway listener's `certificateRefs`, so `ingress.tls.*` has no equivalent and should just be ignored when `httpRoute.enabled`.
- HTTPRoute matches are evaluated most-specific-first per the Gateway API spec, so the `/` catch-all to web can be expressed exactly as it is in the Ingress, with no ordering workarounds.

A related knock-on worth fixing at the same time: `networkPolicy` hardcodes a `podSelector` of `app.kubernetes.io/name: ingress-nginx`, so it must also be disabled on a Gateway API cluster or it black-holes inbound traffic. That is already tracked in #306, but a Gateway API install will hit both together — whatever selector mechanism #306 lands on should cover Gateway implementations too (Envoy Gateway's proxy pods are labelled `app.kubernetes.io/name: envoy` in `envoy-gateway-system`).

### Alternatives Considered

1. **`extraManifests`** — the chart does have this escape hatch, and an HTTPRoute can be pasted in. It still means maintaining the 35-match list by hand, so it solves the mechanics but not the sync problem.
2. **Ingress-to-Gateway translation (`ingress2gateway`)** — works as a one-shot migration, but produces a static snapshot that drifts on the next `helm upgrade`. Also requires leaving `ingress.enabled: true` to have something to convert, which leaves an unreconciled Ingress in the cluster.
3. **Running an ingress-nginx controller alongside the Gateway** — a second north-south data path purely to satisfy one chart. Not viable where the Gateway is the standardised ingress point.

### Component

Helm chart

### Additional Context

Encountered deploying chart 1.9.8 / backend 1.7.1 onto a self-hosted cluster using Envoy Gateway v1.7.1 as the only north-south path (no ingress controller installed at all).

For anyone hitting this before it is supported, this is the equivalent HTTPRoute — `ingress.enabled: false`, backend service port 8080, web 3000:

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: artifact-keeper
  namespace: artifact-keeper
spec:
  parentRefs:
    - group: gateway.networking.k8s.io
      kind: Gateway
      name: envoy
      namespace: classic
      sectionName: artifacts-https
  hostnames:
    - artifacts.example.com
  rules:
    # /api /v2 + the 31 package-format prefixes (PathPrefix),
    # /health /ready (Exact)
    - matches:
        - path: { type: PathPrefix, value: /api }
        - path: { type: PathPrefix, value: /v2 }
        # ... 31 format prefixes omitted for brevity ...
        - path: { type: Exact, value: /health }
        - path: { type: Exact, value: /ready }
      backendRefs:
        - kind: Service
          name: ak-artifact-keeper-backend
          port: 8080
    - matches:
        - path: { type: PathPrefix, value: / }
      backendRefs:
        - kind: Service
          name: ak-artifact-keeper-web
          port: 3000
```

Verified working end to end against chart 1.9.8: web UI, `/health`, `/ready` all serve, and `/v2/` correctly returns 401 from the backend rather than a 404 from the frontend.

### Pre-submission Checklist

damjanek:
## Summary

Fixes #322.

Add opt-in `gateway.networking.k8s.io/v1` HTTPRoute referencing an existing Gateway,
with parent namespace/listener, hostnames, labels, and annotations. Ingress remains
the default; enabling both is rejected. TLS stays on the Gateway. No Gateway,
certificate, or TLS Secret is created.

- Share the 35 existing backend path matches with Ingress: exact health/readiness,
  prefix API/OCI and native package routes. Groups of eight matches support Gateway
  API v1.0.0 CRDs (newer CRDs permit 64).
- Route `/` to enabled web and optionally pass `/dtrack` to enabled Dependency-Track,
  including CORS. Preserve existing Ingress semantics without adding rewriting,
  separate DT frontend, or context-path configuration.
- Add narrowly scoped proxy access with chart NetworkPolicies: explicit namespace
  and nonempty pod-label map required. Existing policies and #306 remain separate.
- Validate configuration; document listener allowedRoutes, TLS, policy requirements,
  and component toggles.
- Bump chart only to 1.10.0, regenerate helm-docs, add CI rendering tests and official
  Gateway API v1.0.0/v1.4.1 schema checks. App version and images stay unchanged.

## Test Checklist
- [ ] Terraform validates/plans cleanly - N/A
- [ ] Manually verified on staging cluster (if applicable) - offline validation only

Rollback: disable `httpRoute.enabled`, restore prior Ingress settings/chart values,
and ensure the Ingress controller is available. HTTPRoute's additive proxy policies
are removed; original policies remain.

Validation:
- 11 unittest methods including 31 malformed-configuration cases, component toggles,
  parent/listener references, metadata, ports, route parity, and policy restrictions.
- Helm 3.16.3/4.1.3; official Gateway v1.0.0/v1.4.1 kubeconform schemas with deliberate
  oversized-match rejection proving maxItems validation is active.
- HTTPRoute + DT/custom-port render: 36/36 resources valid, none skipped.
- Six native lint/template profiles: 143 resources valid, no errors/invalid resources,
  three existing custom resources skipped.
- Default 33-resource and Ingress+DT+edge 36-resource renders unchanged except chart labels.
- helm-docs regeneration, actionlint, diff checks, and chart packaging passed.

No live deployments. Chart version may need reconciliation with other chart PRs.

## Infrastructure
- [ ] Terraform: `terraform validate` passes - N/A
- [ ] Terraform: `terraform plan` shows expected changes - N/A
- [ ] ArgoCD: Application manifests are valid - N/A


brandonrc:
Thanks for this. Checked:
- The HTTPRoute validates against the Gateway API v1.0.0 and v1.4.1 CRDs.
- Its route table matches the Ingress exactly: the same 37 path, service and port entries, from one shared list.
- Enabling it alongside Ingress fails clearly.
- The NetworkPolicy prerequisites fail fast when missing.

Here is what I'll push, keeping your commits:
1. Keep the CORS comment the branch deletes from `dtrack-deployment.yaml`.
2. Add a README note that the Ingress body-size and timeout annotations have no HTTPRoute equivalent.
3. Rebase onto main.

Every chart PR has to bump `Chart.yaml` `version` past main (1.9.12 today), so I will set the final version and regenerate the README with helm-docs 1.14.2 when this reaches the front of the merge queue. Your commits and authorship stay.

PR #338, opened 2026-09-16
fix(chart): opt into encryption keys from existing secrets

### LINKED ISSUE #323: secrets.migrationEncryptionKey / webhookSecretKey act as silent feature toggles when secrets.existingSecret is set
### Description

When `secrets.existingSecret` is set, the chart correctly stops rendering its own Secret and points every workload at the operator-supplied one (#249). But two env vars are still gated on the **chart values** being non-empty rather than on the key existing in the supplied Secret.

`templates/backend-deployment.yaml`:

```
{{- if or (and .Values.externalSecrets.enabled .Values.externalSecrets.secrets.migrationEncryptionKey) (and (not .Values.externalSecrets.enabled) .Values.secrets.migrationEncryptionKey) }}
- name: MIGRATION_ENCRYPTION_KEY
  valueFrom:
    secretKeyRef:
      name: {{ include "artifact-keeper.secretName" . }}
      key: MIGRATION_ENCRYPTION_KEY
{{- end }}
{{- if or (and .Values.externalSecrets.enabled .Values.externalSecrets.secrets.webhookSecretKey) (and (not .Values.externalSecrets.enabled) .Values.secrets.webhookSecretKey) }}
- name: AK_WEBHOOK_SECRET_KEY
  valueFrom:
    secretKeyRef:
      name: {{ include "artifact-keeper.secretName" . }}
      key: AK_WEBHOOK_SECRET_KEY
{{- end }}
```

Note what the condition reads versus what the body references. The guard tests `.Values.secrets.migrationEncryptionKey`, but the value it renders is a `secretKeyRef` into `artifact-keeper.secretName` — which, under `existingSecret`, is the operator's Secret. The value is therefore never used as a credential; it functions purely as a feature flag.

The consequence: an operator who does the recommended thing — put `MIGRATION_ENCRYPTION_KEY` and `AK_WEBHOOK_SECRET_KEY` in their own Secret and leave the chart values empty — gets neither env var injected. The backend then starts with migrations and webhook secret rotation silently disabled, even though both keys are present in the Secret it is already reading `JWT_SECRET` and `DATABASE_URL` from.

This is quiet in a way that is easy to miss. There is no render error, no warning in NOTES.txt, and the pod is `Running` and `Ready`. Per the `values.yaml` documentation the symptoms only appear later, at the point of use:

- migration features: `POST /api/v1/migrations/connections` returns 500
- webhooks: create and rotate-secret are disabled, with only a startup warning

The workaround is to set the values to arbitrary non-empty placeholder strings so the guard passes, while the real keys come from the Secret:

```yaml
secrets:
  existingSecret: my-app-secret
  migrationEncryptionKey: "managed-in-existing-secret"
  webhookSecretKey: "managed-in-existing-secret"
```

That works, but it means putting dummy strings in values.yaml that look like credentials, which is exactly the confusion `existingSecret` is meant to remove. A reviewer seeing that file cannot tell at a glance whether those are real secrets.

### Component

Helm chart

### Environment Information

- Kubernetes Version: v1.31 (self-hosted, kubespray)
- Helm Version: v4.1.3
- Cloud Provider: self-hosted / on-prem
- Secret management: operator-supplied Secret via `secrets.existingSecret` (no External Secrets Operator)

### Chart/Module Version

1.9.8

### Backend Version

1.7.1

### Expected Behavior

When `secrets.existingSecret` is set, both env vars should be injected unconditionally — the chart cannot know which keys the Secret holds, and `secretKeyRef` already fails loudly at pod start if a key is genuinely missing, which is a much better signal than silently dropping a feature.

If opting out needs to stay possible, an explicit pair of booleans would express the intent without overloading credential-shaped values, e.g.:

```yaml
secrets:
  existingSecret: my-app-secret
  migrationEncryptionKeyEnabled: true
  webhookSecretKeyEnabled: true
```

Either way, `values.yaml` should state that under `existingSecret` these entries act as feature toggles rather than credentials — that is the part that is genuinely surprising today.

### Observed Behavior

With `secrets.existingSecret` set and the two values left empty, `MIGRATION_ENCRYPTION_KEY` and `AK_WEBHOOK_SECRET_KEY` are absent from the backend container spec. The pod runs normally; the features are off with no deployment-time signal.

### Steps to Reproduce

```bash
helm template ak ./charts/artifact-keeper \
  --set secrets.existingSecret=my-app-secret \
  | yq '.spec.template.spec.containers[0].env[].name' \
  | grep -E 'MIGRATION_ENCRYPTION_KEY|AK_WEBHOOK_SECRET_KEY'
# -> no output

# Setting the values to any non-empty placeholder injects both:
helm template ak ./charts/artifact-keeper \
  --set secrets.existingSecret=my-app-secret \
  --set secrets.migrationEncryptionKey=x \
  --set secrets.webhookSecretKey=x \
  | yq '.spec.template.spec.containers[0].env[].name' \
  | grep -E 'MIGRATION_ENCRYPTION_KEY|AK_WEBHOOK_SECRET_KEY'
# -> MIGRATION_ENCRYPTION_KEY
# -> AK_WEBHOOK_SECRET_KEY
```

The placeholder value `x` never reaches the container — the env var resolves from `my-app-secret`. That is what confirms it is a pure toggle.

### Relevant Configuration

```yaml
secrets:
  existingSecret: artifact-keeper-db   # holds DATABASE_URL, POSTGRES_PASSWORD,
                                       # JWT_SECRET, DEPENDENCY_TRACK_ADMIN_PASSWORD,
                                       # MIGRATION_ENCRYPTION_KEY, AK_WEBHOOK_SECRET_KEY
  migrationEncryptionKey: ""           # left empty -> env var silently dropped
  webhookSecretKey: ""                 # left empty -> env var silently dropped
```

### Logs

```shell
# Backend starts clean; the only hint is the webhook warning, and there is
# no equivalent for the migration key at startup.
INFO artifact_keeper: HTTP server listening on 0.0.0.0:8080
```

### Additional Context

Found while deploying with `secrets.existingSecret` because `dependencyTrack.enabled` plus `externalDatabase.existingSecret` cannot work together (#150) — consolidating every credential into one operator-managed Secret is the natural way around that, which is what surfaced this.

Both keys were present and correct in the Secret the backend was already using for `JWT_SECRET` and `DATABASE_URL`; only the env var injection was missing.

### Pre-submission Checklist

damjanek:
## Summary

Fixes #323.

Add boolean `secrets.migrationEncryptionKeyEnabled` and
`secrets.webhookSecretKeyEnabled` to reference operator-managed encryption keys
without fake plaintext Helm values. Both default false for compatibility with
Secrets lacking these optional keys. Opt-in requires `secrets.existingSecret` and
disabled ESO; conventional key names become required secretKeyRefs.

Preserve inline, legacy value-based, and ESO provider-path injection. Custom keys
and optional references remain supported through `backend.environmentSecrets`.
Reject conflicting definitions of these two environment variables. No Secret lookup,
plaintext credential rendering, or operator-Secret creation/rotation is introduced.

Document behavior, add offline regressions to CI, regenerate helm-docs 1.14.2, and
bump chart to 1.9.10. Registry and HTTPRoute fixes remain separate; reconcile chart
versions before merge if another PR lands first.

## Test Checklist
- [ ] Terraform validates/plans cleanly - N/A
- [ ] Manually verified on staging cluster (if applicable) - no live deployments

Validation:
- 13 regression methods with parameterized cases on Helm 4.1.3 and 3.16.3: independent
  flags, omitted keys, inline/legacy/ESO, custom/optional references, invalid
  modes/types, conflicts, and disabled backend.
- Original chart fails all three opted-in combinations; the no-flags case passes.
- Helm lint passes on both versions.
- Ten parsed render comparisons identical except version labels: six CI profiles
  and existing/legacy/inline/ESO configurations.
- Strict Kubernetes 1.31 kubeconform: 318 resources, 314 valid, zero invalid/errors,
  four missing CRD schemas skipped.
- Deterministic helm-docs output and clean diff checks.

Rollback: restore previous working values, or equivalent `backend.environmentSecrets`
refs, before downgrading. Old charts ignore the new flags. Operator-managed Secret
contents are unchanged.

## Infrastructure
- [ ] Terraform: `terraform validate` passes - N/A
- [ ] Terraform: `terraform plan` shows expected changes - N/A
- [ ] ArgoCD: Application manifests are valid - N/A


brandonrc:
Thanks for this. Checked against backend main: `MIGRATION_ENCRYPTION_KEY` and `AK_WEBHOOK_SECRET_KEY` are the keys the backend reads.
- With `existingSecret`, nothing is generated or looked up, so the keys stay stable across upgrades.
- No key material reaches a ConfigMap.
- Your 13 tests pass, and every error path fails the way it should.

Here is what I'll push, keeping your commits: rebase onto main, and trim a whitespace-only line the include renders in `backend-deployment.yaml`. `SSO_ENCRYPTION_KEY` isn't set by the chart at all; I'll file that separately. Every chart PR has to bump `Chart.yaml` `version` past main (1.9.12 today), so I will set the final version and regenerate the README with helm-docs 1.14.2 when this reaches the front of the merge queue. Your commits and authorship stay.

PR #339, opened 2026-09-17
feat(chart): make the Dependency-Track API key file optional

### LINKED ISSUE #313: shared-config PVC is RWO and unconditional, so dependencyTrack.enabled pins every backend replica to one node

kalavt:
## Summary

Two self-hosting obstacles from #313, both in `templates/dtrack-deployment.yaml` and the templates around it.

**Replicas cannot spread.** `backend-deployment.yaml` sets `DEPENDENCY_TRACK_API_KEY_FILE` and mounts the `<release>-shared-config` claim whenever `dependencyTrack.enabled` is true. `backend-pvc.yaml` renders that claim from the same single condition. The claim is `ReadWriteOnce` with no `accessModes` knob, so on block storage (EBS, Ceph RBD, Azure Disk) it attaches to one node. Every backend replica is then pinned there. `backend.replicaCount > 1` cannot spread: the second replica fights the first for the volume and stays in `ContainerCreating` with a `Multi-Attach` error. Getting a multi-node backend today costs a shared filesystem nobody wanted.

The volume is not required. `resolve_api_key` in `dependency_track_service.rs` gives `DEPENDENCY_TRACK_API_KEY` precedence over the file. The chart's own comment above that env var points at it. An operator who supplies the key directly still gets the claim attached today.

**The database name is hardcoded.** With `postgres.enabled: false`, `dependency_track` is baked into `ALPINE_DATABASE_URL`. The database must be pre-created under exactly that name. Otherwise a fresh install against an external server crash-loops on `FATAL: database "dependency_track" does not exist` from `UpgradeMetaProcessor`. Two releases sharing one PostgreSQL server collide on it.

## What changed

**1. `dependencyTrack.existingApiKeySecret` / `existingApiKeySecretKey`.** When the name is set, the claim, the backend's mount, `DEPENDENCY_TRACK_API_KEY_FILE` and the Job's mount are all skipped. The backend reads the key from that Secret as `DEPENDENCY_TRACK_API_KEY` instead. One helper carries the condition, so the six call sites cannot drift apart:

```
{{- define "artifact-keeper.dtrackApiKeyFile" -}}
{{- if and .Values.dependencyTrack.enabled (not .Values.dependencyTrack.existingApiKeySecret) -}}true{{- end -}}
{{- end -}}
```

The bootstrap Job still runs. It reconciles the admin password, which DTrack cannot take from the environment. Gating the whole Job on this value would have silently left every new install on the default `admin`/`admin` password. Instead the script takes the path from `DEPENDENCY_TRACK_API_KEY_FILE`, which the Job now sets only in file mode. It exits after the password change when that is unset.

**2. `dependencyTrack.database`,** defaulting to `dependency_track`. Used by `ALPINE_DATABASE_URL` in both the external and the in-chart Postgres branches, and by `CREATE DATABASE` / `GRANT` in `postgres-initdb-configmap.yaml`.

## Notes for review

- **Both defaults render byte-identical manifests.** I diffed a render of `upstream/main` against this branch with the CI's default flags. The only differences are the script's two new guards and the Job's new env var. No claim, mount, env var or JDBC URL moves for existing installs.
- **The explicit value rather than auto-detection.** #313 also proposed detecting `DEPENDENCY_TRACK_API_KEY` in `backend.env` / `backend.environmentSecrets`. I did not do that. `backend.env` may hold a plaintext value and `environmentSecrets` a reference, so a detection rule has to pick a winner between them. Either way it would silently change what an existing install renders, where an explicit value is checkable at render time. I documented the one sharp edge of an explicit value — naming `DEPENDENCY_TRACK_API_KEY` in `backend.env` as well is a duplicate env var, which the API server rejects — the same way the `backend.environmentSecrets` docs already state it.
- **The Job's key extraction now runs only in file mode.** Its `[ -f "$API_KEY_FILE" ]` early exit is guarded with `[ -n ... ]` for the same reason. With no path there is nothing to test, and the Job must still reach the password reconciliation.
- The two changes are separate commits, so they can be reviewed or reverted independently.

## Testing

Rendered with Helm 4.3.0. This repository's CI needs `packages: write` in the org, so what I verified locally:

| Check | Result |
| --- | --- |
| `helm template` default vs. `upstream/main`, CI's flags | identical apart from the two intended hunks |
| `--set dependencyTrack.existingApiKeySecret=ak-dtrack-key` | 5 `PersistentVolumeClaim` objects instead of 6; no `shared-config` in the output except a comment; backend has `DEPENDENCY_TRACK_API_KEY` from `secretKeyRef` and no `/shared` mount |
| `existingApiKeySecretKey` override | renders the given key name |
| `values-production.yaml` with and without the new value | 7 references to `shared-config` vs. 1 |
| `dependencyTrack.database` override, external branch | `jdbc:postgresql://<host>:5432/dtrack_ak` |
| `dependencyTrack.database` override, in-chart Postgres with `postgres.initDb.enabled` | `CREATE DATABASE dtrack_ak;` and `GRANT ALL PRIVILEGES ON DATABASE dtrack_ak TO ...` |
| All six CI variants (`default`, `production`, `staging`, `smoke`, `mesh-main`, `mesh-peer`) | render, exit 0 |
| `helm lint` with the CI's flags | 0 charts failed |
| `helm-docs` 1.14.2 (the CI's version) | run, README committed |

I have not installed this on a cluster with a two-replica backend. The claim is that the volume is gone from the rendered pod spec, not that the replicas spread in a live run.

## Scope

This does not interact with #328, which makes the claim's storage class and access mode configurable. That one relocates the volume to a shared filesystem; this one removes the need for it. Both can land.

I left `Chart.yaml` alone rather than bumping the version, since the release workflow cuts a release on merge.

Closes #313


brandonrc:
Thanks for this. Checked against backend main: `resolve_api_key` reads `DEPENDENCY_TRACK_API_KEY` before `DEPENDENCY_TRACK_API_KEY_FILE`, so the Secret mode works, and existing installs render unchanged.

Here is what I'll push, keeping your commits:
1. Give `dependencyTrack.database` a default of `dependency_track`. An empty value currently renders `CREATE DATABASE ;`.
2. Update the stale note at `values.yaml` ("no way around it today").
3. Carry over #114's guard that skips `CREATE DATABASE` when Dependency-Track shares the main database, credited to its author.
4. Rebase onto main.

This goes first among the Dependency-Track PRs; #110 and #113 follow. Every chart PR has to bump `Chart.yaml` `version` past main (1.9.12 today), so I will set the final version and regenerate the README with helm-docs 1.14.2 when this reaches the front of the merge queue. Your commits and authorship stay.

